PR #186, opened 2022-11-29
Add Puffy hardfork

minh-bq:
The current consensus implementation call WrapUpEpoch before SubmitBlockReward which is the wrong order. We create a hardfork to make the contract call order correct.


DNK90 (approved):
LGTM

linh-1 (approved):

PR #187, opened 2022-11-30
Add reserved gas for system transactions

minh-bq:
Currently, when collecting transactions to commit to a block, we only check the accumulated gas with block's gas limit. Later, when finalizing the block, with system transactions included, the block's gas used may exceed the block's gas limit. As a result, that block is not mined. We want to avoid this scenario by introducing a new flag to reserve some gas in gas pool for system transactions.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -275,6 +275,12 @@ func (w *worker) setGasCeil(ceil uint64) {
 	w.config.GasCeil = ceil
 }
 
+func (w *worker) setGasReserve(reserve uint64) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	w.config.GasReserve = reserve
+}
+
 // setExtra sets the content used to initialize the block extra field.
 func (w *worker) setExtra(extra []byte) {
 	w.mu.Lock()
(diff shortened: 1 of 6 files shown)


linh-1 (approved):

miner/worker.go
```diff
@@ -813,6 +819,12 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 	gasLimit := w.current.header.GasLimit
 	if w.current.gasPool == nil {
 		w.current.gasPool = new(core.GasPool).AddGas(gasLimit)
+
+		// If the gas pool is newly created, reserve some gas for system transactions
+		if err := w.current.gasPool.SubGas(w.config.GasReserve); err != nil {
+			log.Error("Failed to reserve gas for system transactions", "pool", w.current.gasPool, "reserve", w.config.GasReserve)
```
trantienduchn, line 825: should be logging error?
```go
log.Error("Failed to reserve gas for system transactions", "pool", w.current.gasPool, "reserve", w.config.GasReserve, err)
```

miner/worker.go
```diff
@@ -813,6 +819,12 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 	gasLimit := w.current.header.GasLimit
 	if w.current.gasPool == nil {
 		w.current.gasPool = new(core.GasPool).AddGas(gasLimit)
+
+		// If the gas pool is newly created, reserve some gas for system transactions
+		if err := w.current.gasPool.SubGas(w.config.GasReserve); err != nil {
```
trantienduchn, line 824: you are in pessimistic mode

PR #192, opened 2022-12-18
fix: remove wrong validatorLimit local variable

minh-bq:
The len(snap.Validators) is not an invariant, it can be changed in the loop when a new list of validators is loaded. This commit removes that variable and uses len(snap.Validators)/2 + 1 instead.


DNK90 (approved):
LGTM

linh-1 (approved):

PR #193, opened 2022-12-20
feat: verify the block delay on broadcasted blocks

minh-bq:
This commit follows the approach of Ramanujan in BSC which includes the block
delay in block timestamp for verification on the recipients.

The block timestamp is calculated by the formula
- block timestamp = max(parent timestamp + block duration + backoff time,
  current timestamp)

The block duration is 3s in the current configuration. When the validator is
inturn, the backoff time is 0, otherwise, each out of turn validator has a
different delay.

The delay equals to
- delay = wiggle + (random multiplier * wiggle/2)

The random multiplier is in range [1, len(current validators)]. The random seed
is the block number, the recipient can use the same seed to get the same random
result for verification.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -4,13 +4,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"github.com/ethereum/go-ethereum/core"
 	"io"
 	"math/big"
 	"math/rand"
 	"sort"
 	"sync"
 	"time"
+
+	"github.com/ethereum/go-ethereum/core"
 
 	"github.com/common-nighthawk/go-figure"
 
@@ -200,10 +201,6 @@ func (c *Consortium) VerifyHeaderAndParents(chain consensus.ChainHeaderReader, h
 	}
 	number := header.Number.Uint64()
 
-	// Don't waste time checking blocks from the future
-	if header.Time > uint64(time.Now().Unix()) {
-		return consensus.ErrFutureBlock
-	}
 	// Check that the extra-data contains the vanity, validators and signature.
 	if len(header.Extra) < extraVanity {
 		return consortiumCommon.ErrMissingVanity
@@ -282,8 +279,17 @@ func (c *Consortium) verifyCascadingFields(chain consensus.ChainHeaderReader, he
 		return err
 	}
 
+	// Retrieve the snapshot needed to verify this header and cache it
+	snap, err := c.snapshot(chain, number-1, header.ParentHash, parents)
+	if err != nil {
+		return err
+	}
+	if err = c.verifyHeaderTime(header, parent, snap); err != nil {
+		return err
+	}
+
 	// All basic checks passed, verify the seal and return
-	return c.verifySeal(chain, header, parents)
+	return c.verifySeal(chain, header, parents, snap)
 }
 
 // snapshot retrieves the authorization snapshot at a given point in time.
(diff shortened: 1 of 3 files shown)


DNK90 (approved):
LGTM

params/config.go
```diff
@@ -461,6 +461,7 @@ type ChainConfig struct {
 
 	// Puffy hardfork fix the wrong order in system transactions included in a block
 	PuffyBlock *big.Int `json:"puffyBlock,omitempty"` // Puffy switch block (nil = no fork, 0 = already on activated)
+	BubaBlock  *big.Int `json:"bubaBlock,omitempty"`  // Buba switch block (nil = no fork, 0 = already on activated)
```
trantienduchn, line 464: is this mandatory? any dangerous if I don't specify this param in my configuration? (Ops perspective)

consensus/consortium/v2/consortium.go
```diff
@@ -755,15 +806,17 @@ func (c *Consortium) Seal(chain consensus.ChainHeaderReader, block *types.Block,
...
-	delay := time.Unix(int64(header.Time), 0).Sub(time.Now()) // nolint: gosimple
-	if header.Difficulty.Cmp(diffInTurn) != 0 {
-		// It's not our turn explicitly to sign, delay it a bit
-		wiggle := time.Duration(len(snap.Validators)/2+1) * wiggleTime
-		delay += time.Duration(rand.Int63n(int64(wiggle))) + wiggleTime // delay for 0.5s more
-
-		log.Trace("Out-of-turn signing requested", "wiggle", common.PrettyDuration(wiggle))
+	// After the Buba hardfork, the delay is included in header time already
+	delay := time.Until(time.Unix(int64(header.Time), 0))
+	if !c.chainConfig.IsBuba(block.Number()) {
```
trantienduchn, line 811: should it be?
- before Buba: keep the old behavior
- after Buba: new behavior

Here I see you changed the way calculating delay time even when chains don't meet Buba! Can we keep it backward compatible?

minh-bq: Before Buba, the header time does not include delay, after Buba, the delay is included into header time for verification. So the point of this code is if not Buba, calculate the additional delay, otherwise, because the delay is included in header time, just do nothing more and wait until the timestamp in header time.

consensus/consortium/v2/consortium.go
```diff
@@ -462,6 +463,58 @@ func (c *Consortium) verifySeal(chain consensus.ChainHeaderReader, header *types
...
+	// The delayMultiplier is random in range [1, len(validators)]
+	delayMultiplier := make([]int, 0, len(snapshot.Validators))
+	for i := 0; i < len(snapshot.Validators); i++ {
+		delayMultiplier = append(delayMultiplier, i+1)
+	}
+	rand.Shuffle(len(delayMultiplier), func(i, j int) {
+		delayMultiplier[i], delayMultiplier[j] = delayMultiplier[j], delayMultiplier[i]
+	})
+
+	return uint64(int(wiggleTime) + delayMultiplier[distance]*int(wiggleTime)/2)
```
trantienduchn, line 487: I still think two layers of randomize (rand.Shuffle and inside distanceToInturn)  is over complicated. Can you elaborate?

trantienduchn:
```go
	source := rand.NewSource(header.Number.Int64())
	rand := rand.New(source)
	randomNumber := rand.Intn(len(snapshot.Validators) - 1)
```

Is this not enough?

params/config.go
minh-bq: I believe that if this param is not provided in json input, this field is set to `nil` when unmarshalling, if this field is `nil`, it means that `isBuba` always returns false

consensus/consortium/v2/snapshot.go
```diff
@@ -220,6 +220,26 @@ func (s *Snapshot) inturn(validator common.Address) bool {
...
+	validatorIndex := 0
+	for i, val := range validators {
+		if val == validator {
+			validatorIndex = i
+			break
+		}
+	}
+
+	offset := (s.Number + 1) % uint64(len(validators))
+	if validatorIndex > int(offset) {
```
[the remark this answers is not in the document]
minh-bq: `validatorIndex == offset` in case validator is inturn, which returns earlier in `backOffTime`, so this case doesn't happen.
But I see the problem of this code is that in case the validator is not in validators list, `validatorIndex == 0`, so `validatorIndex` can equal to `offset` if `offset == 0` too.

consensus/consortium/v2/consortium.go
trantienduchn: we need a better distribution through every validators delay to minimize block broadcast conflicts, every delay from every node should be different

trantienduchn (approved):

linh-1 (approved):

PR #195, opened 2022-12-30
fix: avoid producing empty block when pending transactions is high

minh-bq:
This commit adapts BSC PR: https://github.com/bnb-chain/bsc/pull/112 to Ronin.

An empty block is pre-sealed by default. It's expected that a new block with
transactions is committed and abort the empty block producer if there are
pending transactions. However, with the low block time, in case there are a lot
of pending transactions, the commit transactions time exceeds the time an empty
block is produced. As a result, block with transactions is not produced but an
empty block and no real transactions is mined.

### DIFF
--- a/cmd/ronin/main.go
+++ b/cmd/ronin/main.go
@@ -131,6 +131,7 @@ var (
 		utils.MinerExtraDataFlag,
 		utils.MinerRecommitIntervalFlag,
 		utils.MinerNoVerifyFlag,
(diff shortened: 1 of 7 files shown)


minh-bq:
This branch is based on https://github.com/axieinfinity/ronin/pull/193 to only support Buba hardfork, this PR is intended to be merged after PR 193

DNK90 (approved):

linh-1 (approved):

PR #197, opened 2023-01-19
fix: improve the bloom filter calculation

minh-bq:
- fix: correct the usage of AsyncReceiptBloomGenerator
AsyncReceiptBloomGenerator allows us the calculating bloom asynchronously with the caller, in this case the caller is applyTransaction. As a result, we don't need to CreateBloom in applyTransaction.
- feat: merge calculated blooms in receipts instead of recalculating
When all receipts have the calculated bloom filters, we can get the final bloom
filter by ORing them instead of looping through logs in receipts and
recalculating all things.
- test: add MergeBloom unit test and benchmark

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -85,7 +85,7 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	}
 	// Validate the received block's bloom with the one derived from the generated receipts.
 	// For valid blocks this should always validate to true.
-	rbloom := types.CreateBloom(receipts)
+	rbloom := types.MergeBloom(receipts)
 	if rbloom != header.Bloom {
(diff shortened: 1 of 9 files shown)


minh-bq:
Benchmark result:

goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/types
BenchmarkCreateBloom/small-8              576574              1870 ns/op             232 B/op          9 allocs/op
BenchmarkCreateBloom/large-8                5888            181090 ns/op           22415 B/op        801 allocs/op
BenchmarkCreateBloom/small-mergebloom-8                  5616206               208.5 ns/op             0 B/op          0 allocs/op
BenchmarkCreateBloom/large-mergebloom-8                    63792             18262 ns/op               0 B/op          0 allocs/op

DNK90 (approved):

PR #205, opened 2023-02-14
fix: duplicate logs returned by StateProcessor.Process

minh-bq:
As we will collect all logs from the receipts after finalizing the block, we must remove the log appending in each transaction to avoid duplication.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -109,7 +109,6 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		if err != nil {
 			return nil, nil, nil, 0, fmt.Errorf("could not apply tx %d [%v]: %w", i, tx.Hash().Hex(), err)
 		}
-		allLogs = append(allLogs, receipt.Logs...)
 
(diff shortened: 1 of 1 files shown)


minh-bq:
AFAIK, luckily, this bug only affects logs subscription API without any impact on chaindata. As a result, no hardfork is needed.

minh-bq:
Before:
```
go test -test.v -run=^TestForkLogsReborn
=== RUN   TestForkLogsReborn
    simulated_test.go:1278: Event should be removed
    simulated_test.go:1291: Event should be included
--- FAIL: TestForkLogsReborn (0.01s)
FAIL
exit status 1
FAIL	github.com/ethereum/go-ethereum/accounts/abi/bind/backends	0.015s
```
After:
```
go test -test.v -run=^TestForkLogsReborn
=== RUN   TestForkLogsReborn
--- PASS: TestForkLogsReborn (0.01s)
PASS
ok  	github.com/ethereum/go-ethereum/accounts/abi/bind/backends	0.016s
```

linh-1 (approved):

PR #220, opened 2023-02-28
fix: add slashed address to verify double sign proof function

minh-bq:
We add a check to ensure the slashed address is matched with the recovered
address. Besides, we add another check to verify the block in proof is not older
than 28800 block from current block.


minh-bq:
This commit is based PR #219, the only new commit is https://github.com/axieinfinity/ronin/pull/220/commits/69297767c260f590cf9423216a5e4b440aaa70bb

DNK90 (approved):

nxqbao (approved):

PR #224, opened 2023-03-01
feat: add limit for block range in log filter

minh-bq:
We add an environment variable to set the block range limit in log filter query.

### DIFF
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -19,7 +19,10 @@ package filters
 import (
 	"context"
 	"errors"
+	"math"
 	"math/big"
+	"os"
+	"strconv"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


eth/filters/filter.go
```diff
@@ -142,6 +161,9 @@ func (f *Filter) Logs(ctx context.Context) ([]*types.Log, error) {
 	if f.end == -1 {
 		end = head
 	}
+	if end-uint64(f.begin) > blockRangeLimit {
+		return nil, errors.New("filter block range is higher than the limit")
```
huyngopt1994, line 165: @minh-bq  Could u print `end-uint64(f.begin)` as well for tracking

minh-bq: Added

DNK90 (approved):

PR #234, opened 2023-03-09
fix: invalid account root traceAPI

linh-1:
## Why
Get the current StateDB instead of using the parent StateDB since the old one is not updated so it will cause the `StateRoot` to be invalid.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -194,10 +194,23 @@ type txTraceResult struct {
 	Error  string      `json:"error,omitempty"`  // Trace failure produced by the tracer
 }
 
+type stateAccount struct {
+	Address     common.Address `json:"address"`
+	Nonce       uint64         `json:"nonce"`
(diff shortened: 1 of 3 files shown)


eth/tracers/api.go
```diff
@@ -745,7 +761,28 @@ func (api *API) traceInternalsAndAccounts(ctx context.Context, block *types.Bloc
...
 
-	results.DirtyAccounts = statedb.DirtyAccounts(blockHash, block.NumberU64())
+	// Get new statedb from the current block to get the latest data
+	newStatedb, err := api.backend.StateAtBlock(ctx, block, reexec, nil, true, false)
+	if err != nil {
+		return nil, err
+	}
+
+	results.DirtyAccounts = make([]*stateAccount, 0)
+	for _, acc := range statedb.DirtyAccounts(blockHash, block.NumberU64()) {
```
minh-bq, line 771: Why do we need this loop?

PR #238, opened 2023-03-13
fix: README misses init genesis instruction

DNK90:
README is currently missing init genesis instruction. Missing genesis block would cause peering issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,15 @@ or, to build the full suite of utilities:
 make all
 ```
 
+### Init genesis
+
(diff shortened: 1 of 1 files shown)


minh-bq (approved):

huyngopt1994:
@DNK90 https://github.com/axieinfinity/ronin/blob/master/docker/chainnode/entrypoint.sh#L60 we had this logic in entrypoint.sh before using this flag FORCE_INIT.

PR #250, opened 2023-03-29
fix: missing reorged blocks

DNK90:
**Purpose:**
After reorg happened, only canonical block is published. Added blocks during this process will not be published.

**How to**
- Add dirtyAccounts and internal transactions caches to cache latest 32 blocks data, write them on insertchain function
- Also add flag to enable storing internal transactions to db if needed (to prevent storage cost) and include writing internal transactions into db after caching internal transactions
- Get these values in reorg function at processing newchain blocks and publish them via event `chainFeed`


cmd/ronin/config.go
```diff
@@ -161,6 +162,9 @@ func makeFullNode(ctx *cli.Context) (*node.Node, ethapi.Backend) {
 	}
 	backend, eth := utils.RegisterEthService(stack, &cfg.Eth)
 
+	// store internal transaction is enabled flag
+	rawdb.WriteStoreInternalTransactionsEnabled(backend.ChainDb(), ctx.GlobalBool(utils.StoreInternalTransactions.Name))
```
minh-bq, line 166: In my opinion, it's better to add a field to blockchain object and read from it to check whether this feature is enabled instead of reading from DB

core/blockchain_reader.go
```diff
@@ -400,6 +400,38 @@ func (bc *BlockChain) SubscribeDirtyAccountEvent(ch chan<- []*types.DirtyStateAc
 	return bc.scope.Track(bc.dirtyAccountFeed.Subscribe(ch))
 }
 
+func (bc *BlockChain) WriteInternalTransactions(hash common.Hash, internalTxs []*types.InternalTransaction) {
+	// cache first
+	bc.internalTransactionsCache.Add(hash, internalTxs)
+
+	// check if store internal transactions is enabled or not
+	if !rawdb.ReadStoreInternalTransactionsEnabled(bc.db) {
```
minh-bq, line 408: I think we should check this before caching the internalTxs

core/blockchain_reader.go
```diff
@@ -400,6 +400,38 @@ func (bc *BlockChain) SubscribeDirtyAccountEvent(ch chan<- []*types.DirtyStateAc
...
+		return
+	}
+	rawdb.WriteInternalTransactions(bc.db, hash, internalTxs)
```
minh-bq, line 411: The internal transactions are written to db and never deleted. I think we should cache them to LRU cache and store them to DB only on blockchain.Stop, load them from DB when block.NewBlockChain. The same logic can apply to dirty accounts.

core/blockchain.go
```diff
@@ -1712,6 +1721,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 
 		if dirtyAccounts != nil {
 			bc.dirtyAccountFeed.Send(dirtyAccounts)
+			bc.dirtyAccountsCache.Add(block.Hash(), dirtyAccounts)
```
minh-bq, line 1724: I think it's better to cache dirty accounts when a configuration is on only, not to always do this.

cmd/ronin/config.go
DNK90: it would be complicated to pass the config to blockchain object from cli and a lot of code changes. Therefore I think, the most simple wait is add it to db

DNK90: I've added a field `shouldStoreInternalTxs` in blockchain object, and get it from db to reduce validation cost everytine `WriteInternalTransactions` is triggered

core/blockchain_reader.go
DNK90: I think if we have the options of storing internal transactions within db. We should accept the increment of the storage. I agree with the logic, storing dirtyAccounts to db  when blockchain is stop

core/blockchain.go
DNK90: I think dirty accounts cache is not big compare to other caches such as block. Therefore, we don't need to add more configuration for it

core/blockchain.go
```diff
@@ -416,9 +416,25 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
...
+
+	// load the latest dirty accounts stored from last stop to cache
+	bc.loadLatestDirtyAccounts()
+
 	return bc, nil
 }
 
+func (bc *BlockChain) loadLatestDirtyAccounts() {
+	for i := uint64(dirtyAccountsCacheLimit); i > 0; i-- {
+		hash := rawdb.ReadCanonicalHash(bc.db, bc.CurrentBlock().NumberU64()-i)
```
minh-bq, line 428: We want to store the dirty accounts of side chain, with this logic, dirty accounts of side chain are not loaded. I think we store them as an array of struct { blockHash, dirtyAccounts }

minh-bq (approved):

PR #252, opened 2023-03-31
feat: reduce block time of out of turn miner

minh-bq:
This commit takes inspiration from section 5.3 BEP-172: Network Stability Enhancement On Slash Occur (https://github.com/bnb-chain/BEPs/pull/172)

When a validator is down, its block turn is mined by an out of turn miner. So when comes the turn of out of turn miner, it may not be able to produce block due to it is already in recently signed list. This leads to a series of out of turn blocks which increase the overall block time because currently out of turn block takes a least 4s. This commit allows out of turn block to take only 3s if the inturn validator is already in recently signed list.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -43,6 +43,7 @@ const (
 
 	validatorBytesLength = common.AddressLength
 	wiggleTime           = 1000 * time.Millisecond // Random delay (per signer) to allow concurrent signers
+	unSealableValidator  = -1
 )
 
 // Consortium delegated proof-of-stake protocol constants.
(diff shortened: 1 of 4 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -491,6 +491,19 @@ func backOffTime(header *types.Header, snapshot *Snapshot) uint64 {
 		return 0
 	}
 
+	var initialDelay time.Duration
+	if chainConfig.IsOlek(new(big.Int).SetUint64(snapshot.Number + 1)) {
+		inturnValidator := snapshot.supposeValidator()
+		pos, _ := snapshot.sealableValidators(inturnValidator)
+		if pos == -1 {
```
DNK90, line 498: make -1 into a constant variable

consensus/consortium/v2/consortium.go
```diff
@@ -491,6 +491,19 @@ func backOffTime(header *types.Header, snapshot *Snapshot) uint64 {
 		return 0
 	}
 
+	var initialDelay time.Duration
```
DNK90, line 494: set `initialDelay = time.Second` here, so we can prevent assigning it in 2 `else` clauses

DNK90 (approved):
LGTM

DNK90 (approved):

PR #257, opened 2023-04-06
fix: password generation inside docker entrypoint

voron:
It seems that password generation is broken for some time, but pr #77 makes it impossible to use any entrypoint args. `${1}` with current password generation implementation is the first entrypoint argument. At the same time we pass all entrypoint's arguments `"$@"` as extra arguments to ronin itself, thus we cannot use password length as an argument


minh-bq (approved):
LGTM
Thanks for your contribution.

PR #258, opened 2023-04-11
chore: reduce log level of snapshot registration failure

minh-bq:
This failure happens when a peer connects with snap protocol but does not support eth protocol. This is not so critical so we reduce the log level from error to info.


DNK90 (approved):
LGTM

PR #259, opened 2023-04-12
fix: handle submitBlockReward correctly when processing system transanctions

minh-bq:
When getting over the DPoS hardfork, we get "insufficient funds for gas * price + value"
error when using debug API.

As the submitBlockReward is a system transaction, it's handled differently from
the normal transactions. When consensus processes submitBlockReward and there
are rewards from transaction fee in block, consensus transfers balance from
SystemAddress to miner address through the statedb.SetBalance directly. Later,
when the system transaction is created, this reward will be transfer from miner
to staking contract.

Currently, in debug API and state accessor, we treat all transactions as normal
transactions.  Without special handling, we miss the state transaction to
transfer balance from SystemAddress to miner. Consequently, when the system
transaction is processed, we get the error because there is a transfer from
miner to staking contract but we see that miner's balance is not sufficient for
the transfer.

This commit checks whether the transaction is submitBlockReward with non-zero
msg.value then adds the state transition to transfer from SystemAddress to miner
to fix the above error.

Fixes: #260

### DIFF
--- a/consensus/consortium/main.go
+++ b/consensus/consortium/main.go
@@ -1,18 +1,20 @@
 package consortium
 
 import (
+	"math/big"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
 	consortiumCommon "github.com/ethereum/go-ethereum/consensus/consortium/common"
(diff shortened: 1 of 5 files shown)


DNK90 (approved):
LGTM

PR #261, opened 2023-04-14
ci: setup public pipline for master branch with nightly tag

huyngopt1994:
We need to setup the other public pipeline for main branch which supports for development phase.

### DIFF
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -1,0 +1,49 @@
+name: master release
+on:
+  push:
+    branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-master.yml
```diff
@@ -0,0 +1,49 @@
+name: master release
+on:
+  push:
+    branches:
+      - main
```
minh-bq, line 5:
```suggestion
      - master
```

chuong-lh (approved):

PR #265, opened 2023-04-19
fix: abort if the private key is different from existed account

minh-bq:
- fix: don't create account automatically

Currently, we automatically create account when no private key is provided and
there is not account in keystore. This behavior may lead to mistakes when
setting up node, so we remove this behavior in this commit.

- fix: abort if the private key is different from existed account

This commit creates a subcommand check in account command of Ronin to check if
the corresponding account of provided private key exist in the imported account.
When starting Ronin from entrypoint.sh, if the provided private key is different
from imported account, the process is aborted.

### DIFF
--- a/cmd/ronin/accountcmd.go
+++ b/cmd/ronin/accountcmd.go
@@ -35,7 +35,7 @@ var (
 		ArgsUsage: "",
 		Category:  "ACCOUNT COMMANDS",
 		Description: `
-    geth wallet import /path/to/my/presale.wallet
+    ronin wallet import /path/to/my/presale.wallet
 
 will prompt for your password and imports your ether presale account.
 It can be used non-interactively with the --password option taking a
@@ -55,7 +55,7 @@ passwordfile as argument containing the wallet password in plaintext.`,
 					utils.LightKDFFlag,
 				},
 				Description: `
-	geth wallet [options] /path/to/my/presale.wallet
+	ronin wallet [options] /path/to/my/presale.wallet
 
 will prompt for your password and imports your ether presale account.
 It can be used non-interactively with the --password option taking a
@@ -111,7 +111,7 @@ Print a short summary of all accounts`,
 					utils.LightKDFFlag,
 				},
 				Description: `
-    geth account new
+    ronin account new
 
 Creates a new account and prints the address.
 
@@ -136,7 +136,7 @@ password to file or expose in any other way.
 					utils.LightKDFFlag,
 				},
 				Description: `
-    geth account update <address>
+    ronin account update <address>
 
 Update an existing account.
 
@@ -148,7 +148,7 @@ format to the newest format or change the password for an account.
 
 For non-interactive use the password can be specified with the --password flag:
 
-    geth account update [options] <address>
+    ronin account update [options] <address>
 
 Since only one password can be given, only format update can be performed,
 changing your password is only possible interactively.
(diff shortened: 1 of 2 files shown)


docker/chainnode/entrypoint.sh
```diff
@@ -81,26 +81,40 @@ accountsCount=$(
...
 
 # private key
-if [[ $accountsCount -le 0 ]]; then
-  echo "No accounts found"
-  if [[ ! -z $PRIVATE_KEY ]]; then
+if [[ ! -z $PRIVATE_KEY ]]; then
+  echo "$PRIVATE_KEY" > ./private_key
+  if [[ $accountsCount -le 0 ]]; then
+    echo "No accounts found"
     echo "Creating account from private key"
```
huyngopt1994, line 88: Should we check mine flag here ?

cmd/ronin/accountcmd.go
```diff
@@ -357,3 +374,27 @@ func accountImport(ctx *cli.Context) error {
...
+	if err != nil {
+		utils.Fatalf("Failed to load the private key: %v", err)
+	}
+	address := crypto.PubkeyToAddress(key.PublicKey)
+
+	stack, _ := makeConfigNode(ctx)
+	for _, wallet := range stack.AccountManager().Wallets() {
+		for _, account := range wallet.Accounts() {
+			if account.Address == address {
+				fmt.Printf("Found account %x\n", address)
```
huyngopt1994, line 393: Should we rename this message for more clear like

huyngopt1994, line 393:
```suggestion
				fmt.Printf("The setting account %x\n is existed in keystore folder", address)
```

cmd/ronin/accountcmd.go
```diff
@@ -357,3 +374,27 @@ func accountImport(ctx *cli.Context) error {
...
+				return nil
+			}
+		}
+	}
+	utils.Fatalf("Account %x not found", address)
```
huyngopt1994, line 398:
```suggestion
	utils.Fatalf("Account %x is conflicted in keystore folder", address)
```

docker/chainnode/entrypoint.sh
```diff
@@ -81,26 +81,40 @@ accountsCount=$(
...
+    exitCode=$?
+    if [[ $exitCode -ne 0 ]]; then
+      echo "An account with different address already exists in $KEYSTORE_DIR"
+      echo "Please consider remove account in keystore" \
+        "or unset PRIVATE_KEY environment variable"
+      exit 1
+    fi
+    set -e
   fi
+  rm ./private_key
```
huyngopt1994, line 107: should add a log for tracking this action ?

minh-bq: I think it's not necessary, this is a temporary file we create in the above code, so just remove it when finishing setting up

cmd/ronin/accountcmd.go
minh-bq: This subcommand is used to check if the account exists, when using the subcommand in bash, we will print a clearer message

docker/chainnode/entrypoint.sh
minh-bq: This is the import account part which is before the actual run of ronin node, so I don't think checking mine flag is needed here

cmd/ronin/accountcmd.go
huyngopt1994: Oke I see the message in below logic

huyngopt1994 (approved):

PR #267, opened 2023-04-24
fix: recently signed check when slashing unavailable validator

minh-bq:
The first validator in snap.Recents is shifted out and allowed to seal the block already. However, when determining if we should slash the validator we mark that first validator as unable to seal block which is incorrect. This loosens the slash check rule. This commit fix this bug and create a separate helper function to check if the validator recently signed a block.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -452,13 +452,8 @@ func (c *Consortium) verifySeal(chain consensus.ChainHeaderReader, header *types
 		return errUnauthorizedValidator
 	}
 
-	for seen, recent := range snap.Recents {
-		if recent == signer {
-			// Signer is among recents, only fail if the current block doesn't shift it out
-			if limit := uint64(len(snap.Validators)/2 + 1); seen > number-limit {
-				return consortiumCommon.ErrRecentlySigned
-			}
-		}
+	if snap.IsRecentlySigned(signer) {
+		return consortiumCommon.ErrRecentlySigned
 	}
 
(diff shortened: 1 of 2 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -602,14 +597,7 @@ func (c *Consortium) processSystemTransactions(chain consensus.ChainHeaderReader
...
 		spoiledVal := snap.supposeValidator()
-		signedRecently := false
-		for _, recent := range snap.Recents {
-			if recent == spoiledVal {
-				signedRecently = true
-				break
-			}
-		}
-		if !signedRecently {
+		if !snap.IsRecentlySigned(spoiledVal) {
```
minh-bq, line 600: This change does not backward compatibility, a hardfork may be needed.

consensus/consortium/v2/consortium.go
```diff
@@ -603,10 +598,14 @@ func (c *Consortium) processSystemTransactions(chain consensus.ChainHeaderReader
 		}
 		spoiledVal := snap.supposeValidator()
 		signedRecently := false
-		for _, recent := range snap.Recents {
-			if recent == spoiledVal {
-				signedRecently = true
-				break
+		if c.chainConfig.IsBuba(header.Number) {
```
minh-bq, line 601: This is just a placeholder for next hardfork, will change to IsOlek when https://github.com/axieinfinity/ronin/pull/252 is merged

DNK90 (approved):

Issue #266, opened 2023-04-24
Validator is slashed even though the node is not down


minh-bq:
#### System information

Ronin version: `2.5.2-stable`

#### Actual behaviour
A validator is slashed for missing inturn block even though its node is still running.

#### Expected behaviour
That validator is not slashed.

#### Log
```
INFO [04-22|17:03:13.005] 🔨 mined potential block                  number=23,452,749 hash=bdfbb5..37ed8a
INFO [04-22|17:03:13.005] Unindexed transactions                   blocks=1   txs=2   tail=21,102,750 elapsed="186.252µs"
INFO [04-22|17:03:13.007] Submitted block reward                   block=23,452,750 amount=0
INFO [04-22|17:03:13.008] Commit new mining work                   number=23,452,750 sealhash=c83398..5347e8 uncles=1 txs=0   gas=64909      fees=0               elapsed=2.375ms
WARN [04-22|17:03:13.008] Block sealing failed                     err="signed recently, must wait for others" height=23,452,750
INFO [04-22|17:03:13.073] Submitted block reward                   block=23,452,749 amount=1,196,040,000,000,000
INFO [04-22|17:03:13.076] Imported new chain segment               blocks=1   txs=2   mgas=0.125  elapsed=4.101ms     mgasps=30.410   number=23,452,749 hash=371125..fdf429 dirty=255.11MiB
INFO [04-22|17:03:13.077] Submitted block reward                   block=23,452,750 amount=0
INFO [04-22|17:03:13.078] Commit new mining work                   number=23,452,750 sealhash=c83398..5347e8 uncles=2 txs=0   gas=64909      fees=0               elapsed=1.885ms
INFO [04-22|17:03:16.008] Submitted block reward                   block=23,452,750 amount=888,280,000,000,000
INFO [04-22|17:03:16.010] Commit new mining work                   number=23,452,750 sealhash=21b289..75632b uncles=2 txs=1   gas=109,323    fees=0.00088828      elapsed=3.831ms
WARN [04-22|17:03:16.010] Block sealing failed                     err="signed recently, must wait for others" height=23,452,750
INFO [04-22|17:03:17.105] Slash validator                          number=23,452,750 spoiled=0x20238eB5643d4D7b7Ab3C30f3bf7B8E2B85cA1e7
INFO [04-22|17:03:17.106] Submitted block reward                   block=23,452,750 amount=11,730,600,000,000,000
[...]

PR #268, opened 2023-04-24
fix: miner does not mine its inturn block

minh-bq:
We have an issue that a miner A is inturn for block 10. At block 9, due to a
node is down, miner A mines the out of turn block 9 and broadcast to the
network. As a result, miner A cannot mine block 10 due to recently signed rule.
However, at block 9, there is another miner B that mines block 9' and this block
is accepted by most of nodes in the network. Then based on block 9', block 10 is
mined by miner C and there is a slash transaction for miner A because on network
view block 9 is mined by miner B not miner A. Miner A receives the chain 9' -
10 and reorgs to that chain. In theory, miner can mine the block 10' which
creates a canonical has greater difficulty. This drops the block 10 and the
slash transaction. However, currently, miner tries to mine block 11 instead of
reproducing new block 10. This leads to miner is slashed even though the node is
running well.

To fix this issue, when mining, we try to go backward on the canonical chain to
find whether we can produce a better chain. If we cannot find way to create a
better chain, we start mining based on the current head block same as current
implementation.

Fixes: https://github.com/axieinfinity/ronin/issues/266

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -912,6 +912,47 @@ func (c *Consortium) readSigner() (common.Address, consortiumCommon.SignerFn, co
 	return c.val, c.signFn, c.signTxFn
 }
 
+// GetBestParentBlock goes backward in the canonical chain to find if the miner can
+// create a chain which has more difficulty than current chain. In case the miner
+// cannot create a better chain, this function returns the head block of current
+// canonical chain.
+func (c *Consortium) GetBestParentBlock(chain *core.BlockChain) (*types.Block, bool) {
+	signer, _, _ := c.readSigner()
+
+	currentBlock := chain.CurrentBlock()
+	block := currentBlock
+	prevBlock := chain.GetBlockByHash(block.ParentHash())
+	diffculty := block.Difficulty().Int64()
+	for diffculty < diffInTurn.Int64() {
+		snap, err := c.snapshot(chain, block.NumberU64()-1, block.ParentHash(), nil)
+		if err != nil {
+			return currentBlock, false
+		}
+		// Miner can create an inturn block which helps the chain to have
+		// greater diffculty
+		if snap.supposeValidator() == signer {
+			inRecent := false
+			for seen, recent := range snap.Recents {
+				if recent == signer {
+					if limit := uint64(len(snap.Validators)/2 + 1); seen > snap.Number+1-limit {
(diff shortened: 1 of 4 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -912,6 +912,47 @@ func (c *Consortium) readSigner() (common.Address, consortiumCommon.SignerFn, co
...
+				if recent == signer {
+					if limit := uint64(len(snap.Validators)/2 + 1); seen > snap.Number+1-limit {
+						inRecent = true
+						break
+					}
+				}
+				if !inRecent {
+					return prevBlock, true
+				}
+			}
```
minh-bq, line 945: This will be refactor to
```golang
if !s.IsRecentlySigned(val) {
    return prevBlock, true
}
```
using the function in PR https://github.com/axieinfinity/ronin/pull/267

consensus/consortium/v2/consortium.go
```diff
@@ -912,6 +912,47 @@ func (c *Consortium) readSigner() (common.Address, consortiumCommon.SignerFn, co
...
+				if recent == signer {
+					if limit := uint64(len(snap.Validators)/2 + 1); seen > snap.Number+1-limit {
+						inRecent = true
+						break
+					}
+				}
+			}
+			if !inRecent {
+				return prevBlock, true
+			}
```
minh-bq, line 945: This will be refactor to

if !s.IsRecentlySigned(val) {
    return prevBlock, true
}
using the function in PR https://github.com/axieinfinity/ronin/pull/267

miner/worker.go
```diff
@@ -1098,10 +1108,18 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
...
 	// sealing in advance without waiting block execution finished.
-	if !noempty && atomic.LoadUint32(&w.noempty) == 0 {
+	if !fastCommit && !noempty && atomic.LoadUint32(&w.noempty) == 0 {
 		w.commit(uncles, nil, false, tstart)
 	}
 
+	// Fast commit an empty block which has block number smaller
+	// than current block in canonical chain but have greater
+	// difficulty
+	if fastCommit {
```
DNK90, line 1118: if you arrange these lines before `1111` you don't need to check `!fastcommit`

DNK90 (approved):
LGTM

PR #270, opened 2023-04-27
eth/tracers: apply latest tracers from ethereum

DNK90:
## Purposes
Our tracers seems outdated since v1.10.3. There are a lots of update from this version. Therefore, I made this PR to update all changes from v1.10.4 to latest tracers PR ([#27031](https://github.com/ethereum/go-ethereum/pull/27031))

## Important Updates
- Upgrade go version from 1.17 to 1.20
- Support `prestate` with `diffMode`, multiplexing tracer (`muxTracer`) and `flatCallTracer` (aka parity style tracer)
- Replace `duktape` to `goja` to handle JS tracer
- Support block overrides in debug_traceCall
- Include `revertReason` to callTracer
- Add `onlyTopCall` and `withLog` options to callTracer
- Replace opcode `0xfe` to `INVALID` opcode


eth/tracers/api.go
```diff
@@ -682,86 +767,50 @@ func (api *API) traceInternalsAndAccounts(ctx context.Context, block *types.Bloc
...
 	}
-	statedb, err := api.backend.StateAtBlock(ctx, parent, reexec, nil, true, false)
+	statedb, release, err := api.backend.StateAtBlock(ctx, parent, reexec, nil, true, false)
 	if err != nil {
 		return nil, err
 	}
+	defer release()
+
+	// Native tracers have low overhead
 	// Execute all the transaction contained within the block concurrently
```
minh-bq, line 777: This change makes traceTx not run concurrently anymore. Currently, we only use this API with call2 tracer which is a native tracer and as the comment says that native tracers have low overhead. So I am not sure about the performance impact here. Furthermore, if we choose to do this way, we should remove the `Execute all the transaction contained within the block concurrently` comment here.

eth/tracers/api.go
```diff
@@ -682,86 +767,50 @@ func (api *API) traceInternalsAndAccounts(ctx context.Context, block *types.Bloc
...
-			}
-		}()
-	}
-	// Feed the transactions into the tracers and return
-	var failed error
 	for i, tx := range txs {
-		// Send the trace task over for execution
-		jobs <- &txTraceTask{statedb: statedb.Copy(), index: i}
-
 		// Generate the next state snapshot fast without tracing
```
minh-bq, line 790: We should remove this comment too

eth/tracers/api.go
DNK90: I got your point, it's a mistake, it should be implemented like `traceBlockParallel`

eth/tracers/api.go
```diff
@@ -682,86 +768,83 @@ func (api *API) traceInternalsAndAccounts(ctx context.Context, block *types.Bloc
...
-		statedb.Finalise(vmenv.ChainConfig().IsEIP158(block.Number()))
+		statedb.Finalise(is158)
 	}
+
 	close(jobs)
 	pend.Wait()
 
-	// Computes the current root hash of the state trie
-	statedb.IntermediateRoot(api.backend.ChainConfig().IsEIP158(block.Number()))
-
```
minh-bq, line 759: We must not remove these lines. Dirty account has a field to store the root of the trie, this `IntermediateRoot` call helps to calculate the root of the trie

DNK90: I see

minh-bq (approved):

PR #273, opened 2023-05-10
chore: add warning when starting a mining node without private key environment

minh-bq:


docker/chainnode/entrypoint.sh
```diff
@@ -99,13 +103,15 @@ if [[ ! -z $PRIVATE_KEY ]]; then
...
       echo "Please consider remove account in keystore" \
-        "or unset PRIVATE_KEY environment variable"
+        "or unset the private key environment variable"
       exit 1
     fi
     set -e
   fi
   rm ./private_key
   unset PRIVATE_KEY
+elif [[ "$mine" = "true" ]]; then
```
huyngopt1994, line 113:
```suggestion
# Bypassing unset private key environment 
```
I think we can add a comment for explaining

docker/chainnode/entrypoint.sh
```diff
@@ -80,6 +80,10 @@ accountsCount=$(
   | wc -l
 )
 
+if [[ ! -z $MINE ]]; then
+  mine="$MINE"
+fi
+
 # private key
```
huyngopt1994, line 87:
```suggestion
# Check when setting private key environment 
```

docker/chainnode/entrypoint.sh
minh-bq: As the log is descriptive below, I don't think it is necessary to add comment here

huyngopt1994 (approved):

PR #275, opened 2023-05-11
ci: checkout repository without submodule when building image

minh-bq:
Currently, the only submodule in Ronin is tests/testdata, this submodule is not necessary for building image. This commit removes submodule fetch when checkout repository for building image to reduce the build time.


minh-bq:
Some numbers:
- The checkout repo without submodule step in test action takes 3s
- The checkout repo with submodule in release master takes 1m30s

PR #277, opened 2023-05-23
chore: optimize sortValidators using pdqsort in golang

DNK90:
## Purpose
Golang from v1.18 has more powerful sort algorithm call pdqsort instead of quicksort. Therefore, I change the current implement (quicksort) to built-in sort which is using pdqsort

### DIFF
--- a/core/vm/consortium_precompiled_contracts.go
+++ b/core/vm/consortium_precompiled_contracts.go
@@ -6,8 +6,8 @@ import (
 	"fmt"
 	"io"
 	"math/big"
-	"math/rand"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


core/vm/consortium_precompiled_contracts.go
```diff
@@ -245,42 +245,41 @@ func (c *consortiumValidatorSorting) Run(input []byte) ([]byte, error) {
...
 
-	for i := range validators {
-		cmp := weights[i].Cmp(weights[right])
-		addrsCmp := big.NewInt(0).SetBytes(validators[i].Bytes()).Cmp(big.NewInt(0).SetBytes(validators[right].Bytes())) > 0
-		if cmp > 0 || (cmp == 0 && addrsCmp) {
-			validators[left], validators[i] = validators[i], validators[left]
-			weights[left], weights[i] = weights[i], weights[left]
-			left++
-		}
+func sortValidators(validators []common.Address, weights []*big.Int) ([]common.Address, []*big.Int) {
```
minh-bq, line 275: I think we don't need to return addresses and balances here, we only change the order of elements in the slice and does not append the slice, so any changes we make to slice inside this function result in the change in caller provided's slice.

minh-bq (approved):

PR #294, opened 2023-06-26
rpc: allow configure WS via passing flags

huyngopt1994:


cmd/utils/flags.go
```diff
@@ -1077,6 +1087,14 @@ func setWS(ctx *cli.Context, cfg *node.Config) {
...
 		cfg.WSPathPrefix = ctx.GlobalString(WSPathPrefixFlag.Name)
 	}
+
+	if ctx.GlobalIsSet(WSReadBufferFlag.Name) {
+		cfg.WSReadBuffer = ctx.GlobalInt(WSReadBufferFlag.Name)
+	}
+
+	if ctx.GlobalIsSet(WSWriteBufferFlag.Name) {
+		cfg.WSWriteBuffer = ctx.GlobalInt(WSWriteBufferFlag.Name)
+	}
```
minh-bq, line 1097: This should change to 
```golang
cfg.WSReadBuffer = ctx.GlobalInt(WSReadBufferFlag.Name)
cfg.WSWriteBuffer = ctx.GlobalInt(WSWriteBufferFlag.Name)
```
The `ctx.GlobalInt` returns default value when that environment variables are not provided. With the current code, I think if the environment variables are not provided, `cfg.WSReadBuffer` and `cfg.WSWriteBuffer` are initialized as 0. Please have some tests to confirm this behavior.

huyngopt1994: Good catch, already self-test, in case we don't pass the flag, it set zero value which breaks the default behaviour, updated.

minh-bq (approved):

PR #298, opened 2023-07-04
feat: implement bls accounts and keymanager

DNK90:


accounts/bls/keymanager.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// IncorrectPasswordErrMsg defines a common error string representing an EIP-2335
+// keystore password was incorrect.
+const IncorrectPasswordErrMsg = "invalid checksum"
+
+type SignRequest struct {
+	PublicKey       []byte `json:"public_key,omitempty"`
+	SigningRoot     []byte `json:"signing_root,omitempty"`
+	SignatureDomain []byte `json:"signature_domain,omitempty"`
```
minh-bq, line 23: This seems unused so we can remove this

PR #301, opened 2023-07-05
feat: implement vote manager logics

DNK90:


core/vote.go
minh-bq, line 1: This file is the same file as core/types/vote.go above

core/types/vote.go
minh-bq, line 1: Can you make this file into a separate PR so we can merge this file before?

core/types/vote.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/ethereum/go-ethereum/crypto/bls"
+	"github.com/pkg/errors"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	BLSPublicKeyLength = 48
+	BLSSignatureLength = 96
```
minh-bq, line 14: These constants are in params/config.go already, so we don't need to redefine them here

core/types/vote.go
```diff
@@ -0,0 +1,89 @@
...
+
+	MaxAttestationExtraLength = 256
```
minh-bq, line 16: Is this variable unused?

core/types/vote.go
```diff
@@ -0,0 +1,89 @@
...
+	hash atomic.Value
+}
+
+// VoteAttestation represents the votes of super majority validators.
+type VoteAttestation struct {
+	VoteAddressSet ValidatorsBitSet // The bitset marks the voted validators.
+	AggSignature   BLSSignature     // The aggregated BLS signature of the voted validators' signatures.
+	Data           *VoteData        // The vote data for fast finality.
+	Extra          []byte           // Reserved for future usage.
+}
```
minh-bq, line 48: Is this struct unused?

core/types/vote.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	BLSPublicKeyLength = 48
+	BLSSignatureLength = 96
+
+	MaxAttestationExtraLength = 256
+)
+
+type BLSPublicKey [BLSPublicKeyLength]byte
+type BLSSignature [BLSSignatureLength]byte
```
minh-bq, line 20: These types are in crypto/bls package already, we shouldn't redefine them here

core/types/vote.go
```diff
@@ -0,0 +1,89 @@
...
+	TargetNumber uint64      // The target block number which validator wants to vote for.
+	TargetHash   common.Hash // The block hash of the target block.
+}
+
+// Hash returns the hash of the vote data.
+func (d *VoteData) Hash() common.Hash { return rlpHash(d) }
+
+// VoteEnvelope represents the vote of a single validator.
+type VoteEnvelope struct {
+	VoteAddress BLSPublicKey // The BLS public key of the validator.
```
minh-bq, line 34: I think this should be renamed to PublicKey, address is not a terminology in BLS signature scheme.

core/types/vote.go
DNK90: these are purely slices not the object itself

core/types/vote.go
DNK90: I think the vote is using bls public key, therefore, remove the `BLS` would make confusion

minh-bq: I mean we should change the name of the variable from VoteAddress to PublicKey

core/types/vote.go
minh-bq: But why don't we use common.PublicKey and common.Signature from crypto/bls/common/interface.go?

DNK90: we also store pubKey and secret key at slice of bytes in keymanager. Therefore we keep these type at byte also

core/vote/vote_pool.go
```diff
@@ -0,0 +1,397 @@
...
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/metrics"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+const (
+	maxCurVoteAmountPerBlock    = 21
```
minh-bq, line 20: This must be greater or equal to 22, not sure if we want more than 22 validators in the future.

core/vote/vote_pool_test.go
```diff
@@ -0,0 +1,430 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
```
minh-bq, line 15: Not sure about the license part here, we may need to remove this

core/vote/vote_pool.go
DNK90: we should move this to config

core/vote/vote_journal.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/tidwall/wal"
+
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/metrics"
+)
+
+const (
+	maxSizeOfRecentEntry    = 512
+	maliciousVoteSlashScope = 256
```
minh-bq, line 16: Look like unused constant

PR #304, opened 2023-07-07
consensus: add fast finality interface

minh-bq:


minh-bq:
Currently, this PR is not compiled successfully, this requires types.VoteEnvelop merge.

PR #307, opened 2023-07-11
eth/protocols: implement finality vote communication via ronin protocol

minh-bq:
ronin protocol is an extension protocol of eth, a peer is required to support eth protocol to run ronin protocol. ronin protocol currently only supports NewVotePacket which contains a list of finality votes.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -46,6 +46,7 @@ import (
 	"github.com/ethereum/go-ethereum/eth/filters"
 	"github.com/ethereum/go-ethereum/eth/gasprice"
 	"github.com/ethereum/go-ethereum/eth/protocols/eth"
+	"github.com/ethereum/go-ethereum/eth/protocols/ronin"
 	"github.com/ethereum/go-ethereum/eth/protocols/snap"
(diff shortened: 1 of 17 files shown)


eth/backend.go
```diff
@@ -590,6 +592,11 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 	if s.config.SnapshotCache > 0 {
 		protos = append(protos, snap.MakeProtocols((*snapHandler)(s.handler), s.snapDialCandidates)...)
 	}
+
+	if s.config.EnableFastFinality {
+		protos = append(protos, ronin.MakeProtocols((*roninHandler)(s.handler))...)
```
DNK90, line 597: should we move checking `EnableFastFinality` inside the protocol? So that we can scale further in the future where we might have other logics that needed to send via `ronin` protocol?

minh-bq: Okay, let me try to decouple ronin protocol and fast finality

minh-bq: I added another flags for configuring ronin protocol

PR #309, opened 2023-07-12
core/vote/vote_pool: protect pool against malicious peer

minh-bq:
This commit limits the number of votes in future queue per peer as the vote in future queue is not fully verified. The check happens before the costly basicVerify which has to verify the BLS signature.

### DIFF
--- a/core/vote/vote_pool.go
+++ b/core/vote/vote_pool.go
@@ -4,8 +4,6 @@ import (
 	"container/heap"
 	"sync"
 
-	mapset "github.com/deckarep/golang-set"
-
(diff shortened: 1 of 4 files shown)


core/vote/vote_pool.go
```diff
@@ -28,11 +26,6 @@ const (
 )
 
 var (
-	localCurVotesCounter    = metrics.NewRegisteredCounter("curVotes/local", nil)
```
DNK90, line 31: why these metrics are removed?

minh-bq: We have gauge metrics for the same type so I remove the counter metrics

PR #311, opened 2023-07-14
vote/vote_pool: use TryRLock in FetchVoteByBlockHash

minh-bq:
The FetchVoteByBlockHash is called by the consensus engine and we don't want this function to block the consensus engine's operations. Currently, FetchVoteByBlockHash uses RLock to wait to acquire the read lock but as the writers are uncontrolled, it can be blocked when there are a lot of spamming writers. This commit makes FetchVoteByBlockHash to use TryRLock to keep polling for the lock n times and returns nil in case it cannot acquire the lock. This helps to bound the max duration a caller must wait on this function.

### DIFF
--- a/core/vote/vote_pool.go
+++ b/core/vote/vote_pool.go
@@ -3,6 +3,7 @@ package vote
 import (
 	"container/heap"
 	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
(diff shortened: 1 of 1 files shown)


core/vote/vote_pool.go
```diff
@@ -23,6 +24,9 @@ const (
 	upperLimitOfVoteBlockNumber = 11 // refer to fetcher.maxUncleDist
 
 	chainHeadChanSize = 10 // chainHeadChanSize is the size of channel listening to ChainHeadEvent.
+
+	fetchCheckFrequency = 100 * time.Millisecond
```
DNK90, line 28: I think 100 ms is too much, it should be smaller. Its turn can be taken over by other locks if the waiting time is too long. In my opinion, we should set it small as 1ms and increase the fetchRetry number

minh-bq: I've changed these constants, my only concern here is if the waiting time is too short, when constantly trying to acquire the lock, we make the cache bouncing between CPUs which degrades the system performance.

PR #313, opened 2023-07-18
ethstats: add coinbase to nodeInfo

DNK90:
# Purpose
In order to serve new feature of ronin-stats (knowing coinbase of connected node), I've added coinbase which is get from eth.Etherbase() function.
Empty coinbase is still allowed in case nodes run without adding any account into account manager

### DIFF
--- a/ethstats/ethstats.go
+++ b/ethstats/ethstats.go
@@ -87,9 +87,10 @@ type Service struct {
 	backend backend
 	engine  consensus.Engine // Consensus engine to retrieve variadic block fields
 
-	node string // Name of the node to display on the monitoring page
-	pass string // Password to authorize access to the monitoring page
-	host string // Remote address of the monitoring service
+	coinbase string
+	node     string // Name of the node to display on the monitoring page
+	pass     string // Password to authorize access to the monitoring page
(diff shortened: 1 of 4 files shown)


ethstats/ethstats.go
```diff
@@ -102,13 +103,14 @@ type Service struct {
...
-//   concurrently.
-//   The Close and WriteControl methods can be called concurrently with all other methods.
+//
+//	Connections support one concurrent reader and one concurrent writer.
+//	Applications are responsible for ensuring that no more than one goroutine calls the write methods
+//	  - NextWriter, SetWriteDeadline, WriteMessage, WriteJSON, EnableWriteCompression, SetCompressionLevel
+//	concurrently and that no more than one goroutine calls the read methods
+//	  - NextReader, SetReadDeadline, ReadMessage, ReadJSON, SetPongHandler, SetPingHandler
+//	concurrently.
+//	The Close and WriteControl methods can be called concurrently with all other methods.
```
minh-bq, line 113: Nit: we should not add this hunk

minh-bq (approved):
LGTM

PR #317, opened 2023-07-20
consortium/v2: fix snapshot save after Shillin

minh-bq:
The blsCommon.PublicKey cannot be un/marshalled to json, this commit converts that field to hex string before marshalling to json.


DNK90 (approved):
LGTM

PR #320, opened 2023-07-21
consortium/v2: add contract interaction to get BLS public key

minh-bq:
This commit adds the contract interaction to get BLS public when preparing checkpoint block. This also creates a getCheckpointValidatorsFromContract function to be used in both Prepare and Finalize the checkpoint block. Besides, a new interface ContractInteraction is created for mock test contract interaction.


DNK90 (approved):
LGTM

PR #337, opened 2023-08-07
consortium/v2: distribute finality reward for the voters in parent block

minh-bq:
Because the finality votes are assembled after we FinalizeAndAssemble the block and the finality reward distribution transaction is in FinalizeAndAssemble, we cannot reward the finality voters in the current block right away. Instead, the reward distribution is delayed until next block. So in the FinalizeAndAssemble, we look at the finality voters in the parent header and distribute reward to them.

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -573,6 +573,7 @@ type ConsortiumV2Contracts struct {
 	RoninValidatorSet common.Address `json:"roninValidatorSet"`
 	SlashIndicator    common.Address `json:"slashIndicator"`
 	ProfileContract   common.Address `json:"profileContract"`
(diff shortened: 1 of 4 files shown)


params/config.go
```diff
@@ -569,11 +570,16 @@ func (c *ChainConfig) String() string {
 		profileContract = c.ConsortiumV2Contracts.ProfileContract
 	}
 
+	finalityTrackingContract := common.HexToAddress("")
+	if c.ConsortiumV2Contracts != nil {
+		profileContract = c.ConsortiumV2Contracts.FinalityTracking
```
DNK90, line 575: is it a typo? should it be finalityTrackingContract?

PR #343, opened 2023-08-16
devp2p: add IP filter and ronin mainnet/testnet filter for nodeset filter

minh-bq:
- devp2p: add IP filter and ronin mainnet/testnet filter for nodeset filter

This commit adds IP or-filter, updates the ronin mainnet and testnet
configuration that are used for nodeset filter. This also allows DNS sign to use
password file instead of entering via prompt.

- script: add script to generate DNS TXT record from list of nodes

### DIFF
--- a/cmd/devp2p/dnscmd.go
+++ b/cmd/devp2p/dnscmd.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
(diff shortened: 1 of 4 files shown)


huyngopt1994:
@minh-bq  For script, should we rewrite have helper function ... https://stackoverflow.com/questions/5474732/how-can-i-add-a-help-method-to-a-shell-script

script/dns.sh
```diff
@@ -0,0 +1,55 @@
...
+# Create an encrypted keyfile.json
+ethkey generate --passwordfile password --privatekey private_key
+
+devp2p discv4 crawl --timeout 1m --bootnodes $BOOTNODES all_nodes.json
+
+mkdir -p $DNS_DIR
+devp2p nodeset filter all_nodes.json -eth-network $NETWORK -ip-list $IP_LIST > $DNS_DIR/nodes.json
+devp2p dns sign $DNS_DIR keyfile.json password --domain $DOMAIN
+devp2p dns to-txt $DNS_DIR $DNS_DIR/txt_record.json
+
```
huyngopt1994, line 52:
```suggestion
echo "Clean up files"
```

huyngopt1994:
@minh-bq  Tested successfully in my env, but maybe we should rename the bash file from **_dns.sh_** to **_gen_dns_txt.sh_** for easily understanding the purpose.

huyngopt1994 (approved):

PR #366, opened 2023-10-11
github/workflows: fix the release workflow

minh-bq:
Currently, when publishing a new release using Github UI, we trigger 3 release workflows. It looks like that 3 action types: created, published, released trigger the workflow. This commit restricts the release workflow to run on published activity type only.

Reference:
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
- https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,7 @@
 name: release
 on:
   release:
+    types: [published]
     tags:
       - v*.*.*
 


huyngopt1994:
@minh-bq  Did u test by yourself?

minh-bq:
No, I didn't.
But look at the workflows that run on the last release in the Actions tab

https://github.com/axieinfinity/ronin/actions/runs/6418172180 -> published activity type
image

https://github.com/axieinfinity/ronin/actions/runs/6418172173 -> released activity type
image

https://github.com/axieinfinity/ronin/actions/runs/6418172163 -> created activity type
image

I did some edits on the description and trigger 1 more workflows
https://github.com/axieinfinity/ronin/actions/runs/6418179378 -> edited activity type
image

So I think restricting the published activity type works just fine.

PR #372, opened 2023-10-26
core: add new sponsored transaction type

minh-bq:


minh-bq:
Some txpool benchmark results:
- All legacy tx benchmarks show no significant differences after the PR
- Txpool insert benchmarks: the result is x2 latency in sponsored tx which is expected because most of the time in insert benchmark is spent to recover the address from signature and the sponsored tx has 2 signatures while legacy tx has only 1
```
BenchmarkPoolBatchInsert100-8                        195           6770191 ns/op
BenchmarkPoolBatchInsert1000-8                        18          59692743 ns/op
BenchmarkPoolBatchInsert10000-8                        2         885204212 ns/op

BenchmarkPoolBatchInsertSponsoredTx100-8              93          11686666 ns/op
BenchmarkPoolBatchInsertSponsoredTx1000-8             10         110871408 ns/op
BenchmarkPoolBatchInsertSponsoredTx10000-8             1        1148534316 ns/op
```
- Txpool demotion benchmark
```
BenchmarkPendingDemotion100-8           14102610                80.28 ns/op
BenchmarkPendingDemotion1000-8          15004642                81.92 ns/op
BenchmarkPendingDemotion10000-8         14364111                79.49 ns/op

BenchmarkPendingSponsoredTxDemotion100-8           36813             30138 ns/op
BenchmarkPendingSponsoredTxDemotion1000-8           3608            360668 ns/op
BenchmarkPendingSponsoredTxDemotion10000-8           136           8353408 ns/op
```
In this benchmark, the transactions are from the same sender but with different payer. As a result, when validating sponsored tx in demotion phase, we need to get balance from `n` payer and 1 sender in comparing with 1 sender when validating legacy tx.

DNK90 (approved):

PR #375, opened 2023-10-31
eth/filters: add finalized block subscription

andicrypt:
* Allow users to subscribe to constantly receive notifications about latest finalized blocks

### DIFF
--- a/core/events.go
+++ b/core/events.go
@@ -18,6 +18,7 @@ package core
 
 import (
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
@@ -42,6 +43,11 @@ type ChainEvent struct {
 	Receipts             types.Receipts
 	FinalizedBlockNumber uint64
 	FinalizedBlockHash   common.Hash
+}
+
+type FinalizedBlockInfo struct {
+	FinalizedBlockNumber hexutil.Uint64      `json:"finalizedBlockNumber"`
+	FinalizedBlockHash   common.Hash `json:"finalizedBlockHash"`
 }
 
 type ChainSideEvent struct {
--- a/eth/filters/filter_system.go
+++ b/eth/filters/filter_system.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -52,6 +53,8 @@ const (
 	PendingTransactionsSubscription
 	// BlocksSubscription queries hashes for blocks that are imported
 	BlocksSubscription
+	// FinalizedBlockSubscription
+	FinalizedBlockSubscription
 	// LastSubscription keeps track of the last index
 	LastIndexSubscription
 )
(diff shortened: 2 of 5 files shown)


eth/filters/filter_system.go
```diff
@@ -351,6 +376,15 @@ func (es *EventSystem) handleTxsEvent(filters filterIndex, ev core.NewTxsEvent)
 	}
 }
 
+func (es *EventSystem) handleFinalizedEvent(filters filterIndex, ev core.ChainEvent) {
+	for _, f := range filters[FinalizedBlockSubscription] {
```
DNK90, line 380: Should we check if the hash and block number is empty before sending it to the subscription?
I saw we have a mark to include or not finalize block in `SendBlockEvent` function. In case, the flag `includeFinalized` is false, they might be empty

core/events.go
```diff
@@ -44,6 +44,11 @@ type ChainEvent struct {
 	FinalizedBlockHash   common.Hash
 }
 
+type Finalizer struct {
```
minh-bq, line 47: I think FinalizedBlockInfo may be better name

eth/filters/filter_system.go
```diff
@@ -351,6 +376,15 @@ func (es *EventSystem) handleTxsEvent(filters filterIndex, ev core.NewTxsEvent)
 	}
 }
 
+func (es *EventSystem) handleFinalizedEvent(filters filterIndex, ev core.ChainEvent) {
```
minh-bq, line 379: Because the following event can have the same finalized block with previous one, we don't want to send a new response in this case. We need a variable to keep track last notified finalized block

eth/filters/filter_system_test.go
```diff
@@ -157,6 +157,56 @@ func (b *testBackend) ServiceFilter(ctx context.Context, session *bloombits.Matc
...
+	chan1 := make(chan *core.Finalizer)
+	sub1 := api.events.SubscribeNewFinalizedBlocks(chan1)
+	go func() { // simulate client
+		i1, i2 := 0, 0
+		for i1 != len(chainEvents) || i2 != len(chainEvents) {
+			select {
+			case f := <-chan0:
+				if chainEvents[i1].FinalizedBlockNumber != f.FinalizedBlockNumber {
+					t.Errorf("sub0 received invalid finalized block number on index %d, want %d, got %d", 
+					i1, chainEvents[i1].FinalizedBlockNumber, f.FinalizedBlockNumber)
```
minh-bq, line 185: Need to add a tab to this line to make it easier to read

eth/filters/filter_system_test.go
```diff
@@ -157,6 +157,56 @@ func (b *testBackend) ServiceFilter(ctx context.Context, session *bloombits.Matc
...
+				}
+				i1++
+			case f := <-chan1:
+				if chainEvents[i2].FinalizedBlockNumber != f.FinalizedBlockNumber {
+					t.Errorf("sub1 received invalid finalized block number on index %d, want %d, got %d", 
+					i2, chainEvents[i2].FinalizedBlockNumber, f.FinalizedBlockNumber)
```
minh-bq, line 191: Need to add a tab to this line to make it easier to read

eth/filters/filter_system.go
andicrypt: Oh this's my mistake, I have put the condition within `handleFinalizedEvent`. Please have a look!

core/events.go
andicrypt: I do agree, fixed!

eth/filters/filter_system.go
andicrypt: Thanks, I fixed it. Please have a look!

eth/filters/filter_system.go
```diff
@@ -351,6 +377,20 @@ func (es *EventSystem) handleTxsEvent(filters filterIndex, ev core.NewTxsEvent)
...
+	if ev.FinalizedBlockNumber == 0 || es.lastFinalized == ev.FinalizedBlockNumber {
```
DNK90, line 381: this condition is not correct, if `ev.FinalizedBlockNumber > 0`, it still passes the condition no matter the second condition meet or not

andicrypt:
**Saigon-testnet:** 
Screenshot 2023-11-01 at 15 17 51

**Note:** Remember turning flag `additionalchainevent.enable`  on.

andicrypt:
**Websocket url:** `""`
**Message:**
```
{"jsonrpc": "2.0","method": "eth_subscribe","params": ["newFinalizedBlocks"],"id": 1}
```

**Response:**
```
{
    "jsonrpc": "2.0",
    "method": "eth_subscription",
    "params": {
        "subscription": "0x4c5cd521106240a12aea37bb57ef666f",
        "result": {
            "finalizedBlockNumber": 21709352,
            "finalizedBlockHash": "0x2dea4504e505f0743d443e6db2f0d270831d33f2917d484cd2f878e44839f416"
        }
    }
}
```

_**Note:**_ Remember turning flag additionalchainevent.enable on.

minh-bq (approved):

PR #378, opened 2023-11-13
consortium: Implement mocking logics for consortium v2

DNK90:
## Purpose
In order to setup and test Ronin nodes locally, we need to do the following steps:
- Config and deploy DPOS contracts on consortium v1
- Add DPOS contracts to genesis file and define forked block number to switch to consortium v2

These steps could affect to demands of developing and testing other features.
Therefore, the ideas of running consortium v2 node without DPOS contracts are come up. 

## Implementation
- In order to do that, we need to implement a mock layer for `ContractInteraction` to mock all of the following function (mock_contract):
```
	GetValidators(blockNumber *big.Int) ([]common.Address, error)
	WrapUpEpoch(opts *ApplyTransactOpts) error
	SubmitBlockReward(opts *ApplyTransactOpts) error
	Slash(opts *ApplyTransactOpts, spoiledValidator common.Address) error
	FinalityReward(opts *ApplyTransactOpts, votedValidators []common.Address) error
	GetBlsPublicKey(blockNumber *big.Int, validator common.Address) (blsCommon.PublicKey, error)
```
- Add global variables `Validators` which contains validators set as well as their public keys which is used for finality vote mocking.
- if `Validators` is not nil, mocked contract is used
- Add `--mock.validators` and `--mock.blspublickeys` flags to add validators and public keys to `Validators` variable

## Running steps:
- Generate key and bls key into your datadir (eg: /node)
```
ronin account new --datadir /node/keystore --password /node/keystore/password
ronin account generatebls --finality.blswalletpath /node/keystore --finality.blspasswordpath /node/keystore/password
```
- Run ronin node with flags `--mock.validators` and `--mock.blspublickeys` (validators and public keys are separated by colon(,))

### DIFF
--- a/cmd/ronin/config.go
+++ b/cmd/ronin/config.go
@@ -20,13 +20,14 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/rawdb"
 	"math/big"
 	"os"
 	"reflect"
 	"unicode"
 
+	"github.com/ethereum/go-ethereum/common"
+	consortiumCommon "github.com/ethereum/go-ethereum/consensus/consortium/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
 	"gopkg.in/urfave/cli.v1"
 
 	"github.com/ethereum/go-ethereum/accounts/external"
@@ -88,10 +89,10 @@ type ethstatsConfig struct {
 }
 
 type gethConfig struct {
-	Eth      ethconfig.Config
-	Node     node.Config
-	Ethstats ethstatsConfig
-	Metrics  metrics.Config
+	Eth        ethconfig.Config
+	Node       node.Config
+	Ethstats   ethstatsConfig
+	Metrics    metrics.Config
 }
 
 func loadConfig(file string, cfg *gethConfig) error {
@@ -151,6 +152,14 @@ func makeConfigNode(ctx *cli.Context) (*node.Node, gethConfig) {
 		cfg.Ethstats.URL = ctx.GlobalString(utils.EthStatsURLFlag.Name)
 	}
 	applyMetricConfig(ctx, &cfg)
+
+	// setup mock config
+	if ctx.GlobalIsSet(utils.MockValidatorsFlag.Name) && ctx.GlobalIsSet(utils.MockBlsPublicKeysFlag.Name) {
+		err = consortiumCommon.SetMockValidators(ctx.GlobalString(utils.MockValidatorsFlag.Name), ctx.GlobalString(utils.MockBlsPublicKeysFlag.Name))
+		if err != nil {
+			utils.Fatalf("failed on create mock validators %v", err)
+		}
+	}
 
 	return stack, cfg
 }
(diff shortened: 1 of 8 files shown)


consensus/consortium/v2/snapshot.go
```diff
@@ -227,33 +230,41 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
...
 
-			isShillin := chain.Config().IsShillin(checkpointHeader.Number)
-			// Get validator set from headers and use that for new validator set
-			extraData, err := finality.DecodeExtra(checkpointHeader.Extra, isShillin)
-			if err != nil {
-				return nil, err
-			}
+			if checkpointHeader.Number.Cmp(common.Big0) == 0 {
+				snap.Validators = make(map[common.Address]struct{})
+				for _, validator := range consortiumCommon.Validators.GetValidators() {
```
minh-bq, line 235: I think it's better to use `c.contract.GetValidators` here

consensus/consortium/v2/consortium.go
```diff
@@ -404,11 +404,19 @@ func (c *Consortium) snapshot(chain consensus.ChainHeaderReader, number uint64,
 			break
 		}
 
+		if number == 0 {
```
minh-bq, line 407: We should add some comments to note that is for mocking/testing only

consensus/consortium/v2/snapshot.go
```diff
@@ -227,33 +230,41 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 				return nil, consensus.ErrUnknownAncestor
 			}
 
-			isShillin := chain.Config().IsShillin(checkpointHeader.Number)
-			// Get validator set from headers and use that for new validator set
-			extraData, err := finality.DecodeExtra(checkpointHeader.Extra, isShillin)
-			if err != nil {
-				return nil, err
-			}
+			if checkpointHeader.Number.Cmp(common.Big0) == 0 {
```
minh-bq, line 233: We should add some comments to note that is for mocking/testing only

consensus/consortium/v2/snapshot.go
DNK90: `c.contract` is not accessible on snapshot, that's why I used `Validators` as global variables

cmd/ronin/config.go
```diff
@@ -152,6 +153,14 @@ func makeConfigNode(ctx *cli.Context) (*node.Node, gethConfig) {
 	}
 	applyMetricConfig(ctx, &cfg)
 
+	// setup mock config
+	if ctx.GlobalIsSet(utils.MockValidatorsFlag.Name) && ctx.GlobalIsSet(utils.MockBlsPublicKeysFlag.Name) {
+		consortiumCommon.Validators, err = consortiumCommon.NewMockValidators(ctx.GlobalString(utils.MockValidatorsFlag.Name), ctx.GlobalString(utils.MockBlsPublicKeysFlag.Name))
```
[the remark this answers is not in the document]
DNK90: That means we need to have a function to get validators? such as `GetValidators()`?
In snapshot, `MockContract` cannot be accessed via consortium. Therefore, I need to set it as a global variable, otherwise, we need to add a method to get it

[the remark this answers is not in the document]
minh-bq: Okay, I got the reason why Validators needs to be public, but I think we can move the consortiumCommon.Validators assignment into the consortiumCommon.NewMockValidators

minh-bq (approved):

PR #388, opened 2023-12-22
entrypoint: move the unlock account logic to right after importing key

minh-bq:
Current code does not re-calculate accountsCount after importing private key. As a result, the unlock account logic is not triggered. This commit moves the unlock logic earlier right after we import the key.

### DIFF
--- a/docker/chainnode/entrypoint.sh
+++ b/docker/chainnode/entrypoint.sh
@@ -132,6 +132,16 @@ if [[ ! -z $PRIVATE_KEY ]]; then
   fi
   rm ./private_key
   unset PRIVATE_KEY
+
+  account=$(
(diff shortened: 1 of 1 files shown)


huyngopt1994:
@minh-bq why don't we plus one accountsCount in logic importing Private key ? and keep the remaining logic ?  in this block code 
```
if [[ ! -z $PRIVATE_KEY ]]; then
  echo "$PRIVATE_KEY" > ./private_key
  if [[ $accountsCount -le 0 ]]; then
    echo "No accounts found"
    echo "Creating account from private key"
    ronin account import ./private_key \
      --datadir $datadir \
      --keystore $KEYSTORE_DIR \
      --password $PASSWORD_FILE
```

huyngopt1994:
approve for skipping minor concerns.

huyngopt1994 (approved):

PR #389, opened 2023-12-25
consortium: enable snap sync on Ronin

minh-bq:
- consortium: avoid reading statedb in block header verification path

In snap sync, we only verify header without processing block transactions so the
statedb is not available. As a result, we need to avoid reading statedb in the
header verification path and must use the information from header instead.

This commit disables signer list verification in extra data of checkpoint block
header in consortium version 1 when snap/fast sync. In consortium version 2, at
the forked block, when getting the list of validators, we read this information
from header instead of contract, it is expected to behave the same as before.
This commit does not change the header verification process when full sync.

- eth/protocols/snap: fix problems due to idle-but-busy peers

Mark the peer as idle only when it finish processing response packet

- eth/protocols/snap: throttle trie heal requests when peers DoS us

Throttle the breadth expansion to prioritize depth expansion

- eth/protocols/snap: sort trienode heal requests by path

Sort the heal requests by path to merge the storage trie requests from the same
account. This helps to reduce the bandwith.

- eth/ethconfig: make full sync the default sync mode

Make full sync the default sync mode until we thoroughly test the snap sync.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -337,6 +337,11 @@ func (c *Consortium) VerifyHeaderAndParents(chain consensus.ChainHeaderReader, h
 // in a batch of parents (ascending order) to avoid looking those up from the
 // database. This is useful for concurrently verifying a batch of new headers.
 func (c *Consortium) verifyCascadingFields(chain consensus.ChainHeaderReader, header *types.Header, parents []*types.Header) error {
+	var (
+		snap *Snapshot
+		err  error
+	)
+
 	// The genesis block is the always valid dead-end
 	number := header.Number.Uint64()
 	if number == 0 {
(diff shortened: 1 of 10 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -408,6 +416,53 @@ func (c *Consortium) verifyCascadingFields(chain consensus.ChainHeaderReader, he
...
+	}
+
+	for _, validator := range extraData.CheckpointValidators {
+		validators = append(validators, validator.Address)
+	}
+
+	snap = newSnapshot(c.chainConfig, c.config, c.signatures, number, hash, validators, nil, c.ethAPI)
+
+	// load v1 recent list to prevent recent producing-block-validators produce block again
+	snapV1 := c.v1.GetSnapshot(chain, number, hash, parents)
```
huyngopt1994, line 450: https://github.com/minh-bq/ronin/blob/0a699bc486bfeae669feaa3ec8e35b431206229d/consensus/consortium/v1/consortium.go#L711 seem the error of GetSnapshot is swallowed here ? If any error occurs here, we only receive message snapshot V1 is not available ? Could be return err from GetSnapshot function and return here ?

minh-bq: This code is executed only once at the Consortium block which is in the past, so I don't think adding error here is necessary anymore.

huyngopt1994 (approved):

PR #392, opened 2024-01-09
internal/ethapi: cherry pick EstimateGas improvement from go-ethereum

minh-bq:
Some notable changes:
- EstimateGas on latest block instead of pending block when block argument is not provided
- Add support for block overrides in eth_call
- Add support for state overrides in eth_estimateGas
- Reduce the binary search iterations in EstimateGas
  + Return 21000 for plain transfer when simulated successfully
  + Favor lower bound in binary search
  + Don't try to perfect gas estimation, return early when the lower-higher range is smaller than error ratio
  + Run the second iteration with `gas = (result.UsedGas + result.RefundedGas + params.CallStipend) * 64 / 63` to reduce the binary search range

### DIFF
--- a/consensus/consortium/common/contract.go
+++ b/consensus/consortium/common/contract.go
@@ -437,7 +437,7 @@ func (b *ConsortiumBackend) CallContract(ctx context.Context, call ethereum.Call
 		Gas:  &gas,
 		To:   call.To,
 		Data: &data,
-	}, block, nil)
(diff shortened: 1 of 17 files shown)


DNK90 (approved):

huyngopt1994:
https://github.com/ethereum/go-ethereum/pull/24363

huyngopt1994 (approved):

PR #395, opened 2024-01-11
eth/filters: reduce database read when handling eth_getLogs

minh-bq:
While profiling, we see that when bloom filter hits, 1/3 CPU time in check log
match routine is used to get the corresponding block number from block hash.
This is redundant as this information is available already and can be passed via
function argument.


minh-bq:
Before this PR, around 1/3 CPU time when GetLogs is used to rawdb.ReadHeaderNumber, the other is used to rawdb.ReadLogs
get_logs

After this PR, most of CPU time when GetLogs is used to rawdb.ReadLogs
get_logs_v2

PR #397, opened 2024-01-15
consortium: fix not validated snapshot query

NganSM:
The `verifyFinalitySignatures` function is invoked in `VerifyHeaderAndParents` before the condition
`parent.Number.Uint64() == number-1 && parent.Hash() == header.ParentHash` is checked. This can result in generating and storing an invalid snapshot because the block number and block hash provided as inputs for `snapshot` do not match. If the `if number == c.forkedBlock-1` block is executed, this can lead to overwriting past snapshot with invalid data.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -286,33 +286,6 @@ func (c *Consortium) VerifyHeaderAndParents(chain consensus.ChainHeaderReader, h
 	if header.Number == nil {
 		return consortiumCommon.ErrUnknownBlock
 	}
-	number := header.Number.Uint64()
-
-	isShillin := c.chainConfig.IsShillin(header.Number)
-	extraData, err := finality.DecodeExtra(header.Extra, isShillin)
-	if err != nil {
-		return err
-	}
-
-	// Check extra data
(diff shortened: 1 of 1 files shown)


DNK90 (approved):
LGTM

consensus/consortium/v2/consortium.go
```diff
@@ -358,6 +331,33 @@ func (c *Consortium) verifyCascadingFields(chain consensus.ChainHeaderReader, he
...
+
+	// Check extra data
+	isEpoch := number%c.config.EpochV2 == 0 || c.chainConfig.IsOnConsortiumV2(header.Number)
+
+	if !isEpoch && len(extraData.CheckpointValidators) != 0 {
+		return consortiumCommon.ErrExtraValidators
+	}
+
+	if isShillin && extraData.HasFinalityVote == 1 {
+		if err := c.verifyFinalitySignatures(
```
andicrypt, line 349: For me, it looks good to move this verification here; yet we call `snapshot` twice including one in `verifyCascadingFields` one in line 376, should we better this?

minh-bq: This could be implemented in another PR

minh-bq (approved):

PR #399, opened 2024-01-26
fix: Add Ena hardfork to correct payer hash

NganSM:
The way `payerHash` is calculated (see `transaction_signing.go/payerInternal`) can lead to RLP encoding collision. We should add a prefix before the RLP-encoded data to avoid this.

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -47,6 +47,8 @@ func MakeSigner(config *params.ChainConfig, blockNumber *big.Int) Signer {
 		signer = NewLondonSigner(config.ChainID)
 	case config.IsBerlin(blockNumber):
 		signer = NewEIP2930Signer(config.ChainID)
+	case config.IsEna(blockNumber):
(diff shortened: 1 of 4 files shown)


core/types/transaction_signing.go
```diff
@@ -514,6 +543,117 @@ func (s MikoSigner) Payer(tx *Transaction) (common.Address, error) {
 	return payerInternal(s, tx)
 }
 
+type EnaSigner struct {
+	EIP155Signer
```
minh-bq, line 547: I think making EnaSigner "inherits" MikoSigner avoids duplicated Sender, SignatureValues, Hash code, we just need to override Equal and Payer

params/config.go
minh-bq, line 1: EnaBlock should be added to chainConfigFmt  and CheckCompatible function

minh-bq (commented):
Please update the commit message to follow the convention of other commits (the prefix should be the part of project your change is related to)

PR #407, opened 2024-02-26
rawdb: add Consortium snapshot to inspect db's result

minh-bq:
This commit adds the key that Consortium snapshot uses to store the value into key-value DB to database inspect command. This helps to get storage stats of Consortium snapshot.


andicrypt (approved):
LGTM

PR #409, opened 2024-02-27
docker: make ports configurable

jonreiter:


minh-bq (approved):
Thanks for your contribution. LGTM.

docker/chainnode/entrypoint.sh
```diff
@@ -316,17 +334,17 @@ exec ronin $params \
   --syncmode $syncmode \
   --verbosity $VERBOSITY \
   --datadir $datadir \
-  --port 30303 \
+  --port $NETWORK_PORT \
```
huyngopt1994, line 337: Should use this one $network_port ?  Otherwise the logic is broken when `NETWORK_PORT` is not set ? 
Maybe u can use the other approach like this which is quite easier 
```
--port  ${NETWORK_PORT:-30303}
```

jonreiter: fixed for all 3 new params

PR #414, opened 2024-03-11
core: adjust reserved gas for system transactions in non-checkpoint block

andicrypt:
Currently, reserved gas for system transaction is hardcoded as 10M gas. (https://github.com/axieinfinity/ronin/blob/95e3707865263a9c751307c951ef5c02a8247717/miner/worker.go#L879-L890). However, only the checkpoint block - 1 (the block at the end of epoch) needs to reserve that much (because wrap up epoch transaction requires a lot of gas), it is necessary to adjust the logic to reserve lower gas in normal block or higher gas in checkpoint block.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -872,17 +872,25 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 		return true
 	}
 
+	var reservedGas uint64 = 0
+	if w.chainConfig.Consortium != nil {
+		if w.current.header.Number.Uint64()%w.chainConfig.Consortium.EpochV2 == w.chainConfig.Consortium.EpochV2-1 {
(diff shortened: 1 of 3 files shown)


miner/worker.go
```diff
@@ -872,17 +872,23 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 		return true
 	}
 
+	var reservedGas uint64 = 0
+	if w.current.header.Number.Uint64()%w.chainConfig.Consortium.EpochV2 == w.chainConfig.Consortium.EpochV2-1 {
```
minh-bq, line 876: We need to add check for `pool.chainconfig.Consortium != nil` to avoid crash on unit test

minh-bq (commented):
Please change the prefix in commit title to follow the convention

minh-bq (approved):

PR #419, opened 2024-03-14
consortium-v2, precompiled: add contract for sorting validator with random beacon

minh-bq:
- precompiled: add contract for sorting validator with random beacon

This new precompiled contract provides requestSortValidatorSet and
pickValidatorSet function. The requestSortValidatorSet receives the list of all
validator candidates, pick non-rotating validators based on staked amounts and
rotating validators based on weight derived from random beacon. The list of
validators for all epoch of that period is stored in precompiled contract's
storage. The pickValidatorSet supports query the validator set of an epoch in
that period.

- consortium-v2: enable pick validator with beacon at Tripp hardfork - 1

Since precompiled contract has no code, it can be marked as empty account if
nonce and balance is 0 too. If the account is empty, the account node is deleted
so writing to storage of that account has no effect. We avoid this by setting
nonce of precompiled contract to 1 at Tripp hardfork block - 1 before the
WrapUpEpoch system transaction.

### DIFF
--- a/core/vm/consortium_precompiled_contracts.go
+++ b/core/vm/consortium_precompiled_contracts.go
@@ -1094,7 +1094,7 @@ func pickNonRotatingValidator(
 
 func (contract *pickValidatorSetBeacon) requestSortValidator(method *abi.Method, args []interface{}) ([]byte, error) {
 	if len(args) != 8 {
-		return nil, fmt.Errorf("invalid arguments, expected 4 got %d", len(args))
+		return nil, fmt.Errorf("invalid arguments, expected 8 got %d", len(args))
 	}
 	beacon, ok := args[0].(*big.Int)
 	if !ok {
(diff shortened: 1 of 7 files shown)


core/vm/consortium_precompiled_contracts.go
```diff
@@ -773,3 +754,466 @@ func (contract *consortiumValidateProofOfPossession) Run(input []byte) ([]byte,
...
+	for _, validator := range standardValidator {
+		nonRotatingValidator = append(nonRotatingValidator, validator.address)
+	}
+
+	return nonRotatingValidator
+}
+
+func (contract *pickValidatorSetBeacon) requestSortValidator(method *abi.Method, args []interface{}) ([]byte, error) {
+	if len(args) != 8 {
+		return nil, fmt.Errorf("invalid arguments, expected 4 got %d", len(args))
```
andicrypt, line 1097: minor fix comment, expected 8

andicrypt (approved):

andicrypt:
LSTM

PR #427, opened 2024-03-25
consortium-v2: update voting process logic

andicrypt:
Currently, only block producers can participate in voting process. However, after Tripp, vote permission is granted to a wider range of validator candidates (up to 64). Therefore, this PR differentiates block producers from validator candidates and enables logic for wider voter set. Moreover, voter weight is also introduced in order to determine reward distribution for every eligible candidates. Voter weight is computed based on candidates's staked amounts, according to the formula represented in: https://github.com/phuctd95/REPs/blob/main/REP-0010/REP-0010.md.

Additionally, after Tripp, the logic for getting validator set is also changed such that: 
* At the start of period, the first checkpoint block after 00:00 UTC, the list of validator candidates will be get from the contract once for whole period (~1 day).
* At the start of epoch, the list of block producers will be get from the contract once for every epoch (~10 minutes).

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -274,14 +274,10 @@ func (c *Consortium) verifyFinalitySignatures(
 	parentHash common.Hash,
 	parents []*types.Header,
 ) error {
+	isTripp := c.chainConfig.IsTripp(new(big.Int).SetUint64(parentNumber + 1))
 	snap, err := c.snapshot(chain, parentNumber, parentHash, parents)
 	if err != nil {
 		return err
-	}
-
-	votedValidatorPositions := finalityVotedValidators.Indices()
-	if len(votedValidatorPositions) < int(math.Floor(finalityRatio*float64(len(snap.ValidatorsWithBlsPub))))+1 {
-		return finality.ErrNotEnoughFinalityVote
 	}
 
 	voteData := types.VoteData{
@@ -291,13 +287,34 @@ func (c *Consortium) verifyFinalitySignatures(
 	digest := voteData.Hash()
 
 	// verify aggregated signature
-	var publicKeys []blsCommon.PublicKey
+	var (
+		publicKeys            []blsCommon.PublicKey
+		accumulatedVoteWeight int
+		finalityThreshold     int
+	)
+	votedValidatorPositions := finalityVotedValidators.Indices()
 	for _, position := range votedValidatorPositions {
 		if position >= len(snap.ValidatorsWithBlsPub) {
 			return finality.ErrInvalidFinalityVotedBitSet
 		}
 		publicKeys = append(publicKeys, snap.ValidatorsWithBlsPub[position].BlsPublicKey)
+		if isTripp {
+			accumulatedVoteWeight += int(snap.FinalityVoteWeight[position])
+		} else {
+			accumulatedVoteWeight += 1
+		}
 	}
+
+	if isTripp {
+		finalityThreshold = int(math.Floor(finalityRatio*float64(consortiumCommon.MaxFinalityVotePercentage))) + 1
+	} else {
+		finalityThreshold = int(math.Floor(finalityRatio*float64(len(snap.ValidatorsWithBlsPub)))) + 1
+	}
+
+	if accumulatedVoteWeight < finalityThreshold {
+		return finality.ErrNotEnoughFinalityVote
+	}
+
 	if !finalitySignatures.FastAggregateVerify(publicKeys, digest) {
 		return finality.ErrFinalitySignatureVerificationFailed
 	}
@@ -682,7 +699,10 @@ func (c *Consortium) getCheckpointValidatorsFromContract(
 	)
 
 	isShillin := c.chainConfig.IsShillin(header.Number)
-	if isShillin {
+	isTripp := c.chainConfig.IsTripp(header.Number)
+	// After Tripp, BLS key of validator is read at the start of new period,
+	// not the start of epoch anymore.
+	if isShillin && !isTripp {
 		// The filteredValidators shares the same underlying array with newValidators
 		// See more: https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
 		filteredValidators = filteredValidators[:0]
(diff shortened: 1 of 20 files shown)


consensus/consortium/v2/finality/consortium_header.go
```diff
@@ -305,10 +307,16 @@ func (extraData *HeaderExtraData) EncodeRLP() ([]byte, error) {
...
 			Address:      val.Address,
 			BlsPublicKey: val.BlsPublicKey.Marshal(),
+			StakedAmount: val.StakedAmount,
 		}
 	}
 	ext.CheckpointValidators = cp
 
+	if extraData.BlockProducers != nil {
+		ext.BlockProducers = make([]common.Address, len(extraData.BlockProducers))
+		copy(ext.BlockProducers, extraData.BlockProducers)
```
minh-bq, line 317: I think we don't need deep copy here, isn't `ext.BlockProducers = extraData.BlockProducers` necessary here?

consensus/consortium/v2/finality/consortium_header.go
```diff
@@ -342,9 +350,15 @@ func DecodeExtraRLP(enc []byte) (*HeaderExtraData, error) {
 		cp[i] = ValidatorWithBlsPub{
 			Address:      val.Address,
 			BlsPublicKey: blsPublicKey,
+			StakedAmount: val.StakedAmount,
 		}
 	}
 	ret.CheckpointValidators = cp
+	if dec.BlockProducers != nil {
```
minh-bq, line 357: Same as the above comment

consensus/consortium/v2/snapshot.go
```diff
@@ -232,6 +234,12 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 			}
 		}
 
+		isTripp := chain.Config().IsTripp(header.Number)
+		// Update the current period for every Tripp blocks
+		if isTripp {
+			snap.CurrentPeriod = header.Time / dateInSeconds
```
minh-bq, line 240: This must be updated only at checkpoint block?

consensus/consortium/v2/consortium.go
```diff
@@ -664,28 +697,67 @@ func (c *Consortium) verifyHeaderTime(header, parent *types.Header, snapshot *Sn
...
+			blsPublicKey, err := contract.GetBlsPublicKey(parentBlockNumber, candidate)
+			if err == nil {
+				checkpointValidators = append(checkpointValidators, finality.ValidatorWithBlsPub{
+					Address:      candidate,
+					BlsPublicKey: blsPublicKey,
+					Weight:       weights[i],
+				})
+			}
+		}
+		return checkpointValidators, blockProducers, nil
```
minh-bq, line 748: Do we need to sort checkpointValidators?

consensus/consortium/v2/consortium.go
```diff
@@ -899,18 +980,47 @@ func (c *Consortium) Finalize(chain consensus.ChainHeaderReader, header *types.H
...
 
-			if isShillin {
-				if !checkpointValidator[i].BlsPublicKey.Equals(extraData.CheckpointValidators[i].BlsPublicKey) {
-					return errMismatchingEpochValidators
+			for i := range checkpointValidators {
+				if checkpointValidators[i].Address != extraData.CheckpointValidators[i].Address {
+					return errMismatchingValidators
+				}
+
+				if c.chainConfig.IsShillin(header.Number) {
```
minh-bq, line 1020: As this does not change during iteration, I think it's better to move this outside of the loop.

consensus/consortium/v2/consortium.go
```diff
@@ -1461,3 +1595,60 @@ func encodeSigHeader(w io.Writer, header *types.Header, chainId *big.Int) {
...
+			parentNumber uint64        = number - 1
+			parentHash   common.Hash   = header.ParentHash
+			parent       *types.Header = chain.GetHeader(parentHash, parentNumber)
+		)
+		for parentNumber%c.config.EpochV2 != 0 && parent != nil {
+			parentNumber = parent.Number.Uint64() - 1
+			parentHash = parent.ParentHash
+			parent = chain.GetHeader(parentHash, parentNumber)
+		}
+		return uint64(header.Time/dateInSeconds) > uint64(parent.Time/dateInSeconds)
```
minh-bq, line 1622: The above loop can break if parent == nil. In that case, parent.Time results into a nil pointer dereference.

minh-bq: fixed

consensus/consortium/v2/consortium.go
```diff
@@ -1620,3 +1641,34 @@ func (c *Consortium) IsPeriodBlock(chain consensus.ChainHeaderReader, header *ty
...
+
+		snap, err := c.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, nil)
+		if err != nil {
+			log.Error("Failed to get snapshot", "err", err)
+			parent := c.getLastCheckpointHeader(chain, header)
+			if parent == nil {
+				return false
+			}
+
+			return uint64(header.Time/dateInSeconds) > uint64(parent.Time/dateInSeconds)
```
andicrypt, line 1661: I think this should be `parent.Time/dateInSeconds > c.chainConfig.TrippPeriod`

consensus/consortium/v2/consortium.go
```diff
@@ -1620,3 +1641,34 @@ func (c *Consortium) IsPeriodBlock(chain consensus.ChainHeaderReader, header *ty
...
+		}
+
+		if snap.CurrentPeriod > c.chainConfig.TrippPeriod.Uint64() {
+			return true
+		}
+
+		if c.IsPeriodBlock(chain, header) {
```
andicrypt, line 1668: I think we can check like `header.Number%EpochV2==0 && header.Time/dateInSeconds>c.chainConfig.TrippPeriod`, this can reduce a `snapshot` in IsPeriodBlock

consensus/consortium/v2/consortium.go
minh-bq: fixed

consensus/consortium/v2/consortium.go
minh-bq: fixed

PR #431, opened 2024-04-03
consortium-v2: fix the EOA check of system transaction

minh-bq:
The EOA check wants to make sure that the sender of system transaction has no code. The code hash of existing account that has no code is crypto.Keccak256Hash(nil). Meanwhile, an empty/non-existing account's code hash is 0x000...00. Currently, we miss the case for empty account, which makes the system transaction from empty account be rejected.


NganSM:
I think the check is a bit redundant since the probability that someone can deploy a contract in place of an EOA is negligible due to the way a contract's address is calculated. The transaction's signature has already been validated in `IsSystemTransaction`, so a system transaction is most likely to be from an EOA.

DNK90 (approved):

PR #435, opened 2024-04-04
params/bootnodes: remove default etherum mainnnet bootnodes

huyngopt1994:
Replace the default Ethereum foundation mainnet bootnodes by Ronin bootnodes for avoiding unused discovery and peering in Network.  We have 2 options for extending the bootnodes now 
1. Overwrite by setting Bootnodes ENV
2. We automatically fetch 5 endpoints from DNS discovery, implemented in https://github.com/axieinfinity/ronin/pull/370

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1000,12 +1000,12 @@ func setNodeUserIdent(ctx *cli.Context, cfg *node.Config) {
 // setBootstrapNodes creates a list of bootstrap nodes from the command line
 // flags, reverting to pre-configured ones if none have been specified.
 func setBootstrapNodes(ctx *cli.Context, cfg *p2p.Config) {
-	urls := params.EtherumMainnetBootnodes
+	urls := params.RoninMainnetBootnodes
 	switch {
 	case ctx.GlobalIsSet(BootnodesFlag.Name):
 		urls = SplitAndTrim(ctx.GlobalString(BootnodesFlag.Name))
 	case ctx.GlobalBool(MainnetFlag.Name):
-		urls = params.RoninMainnetBootnodes
+		urls = params.EtherumMainnetBootnodes
 	case ctx.GlobalBool(RopstenFlag.Name):
 		urls = params.RopstenBootnodes
 	case ctx.GlobalBool(SepoliaFlag.Name):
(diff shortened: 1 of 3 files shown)


cmd/utils/flags.go
```diff
@@ -1000,10 +1000,12 @@ func setNodeUserIdent(ctx *cli.Context, cfg *node.Config) {
 // setBootstrapNodes creates a list of bootstrap nodes from the command line
 // flags, reverting to pre-configured ones if none have been specified.
 func setBootstrapNodes(ctx *cli.Context, cfg *p2p.Config) {
-	urls := params.MainnetBootnodes
+	urls := params.EtherumMainnetBootnodes
 	switch {
 	case ctx.GlobalIsSet(BootnodesFlag.Name):
 		urls = SplitAndTrim(ctx.GlobalString(BootnodesFlag.Name))
+	case ctx.GlobalBool(MainnetFlag.Name):
```
minh-bq, line 1007: We never set this flag when running, so this looks incorrect. The default must be `RoninMainnetBootnodes` and in this case when mainnet flag is set `urls = params.EtherumMainnetBootnodes`

huyngopt1994: Yes, checked this flag, it's used for Etherum Mainnet. I have the otherway for creating flag for Ronin but it requires more flag and updating in entrypoint.sh, so let go with this suggestion.

DNK90:
I think we should remove all ethereum network. Just add 2 types: our saigon testnet and mainnet.
It would prevent misconfiguration from users

mobile/params.go
```diff
@@ -71,8 +71,8 @@ func GoerliGenesis() string {
 // FoundationBootnodes returns the enode URLs of the P2P bootstrap nodes operated
 // by the foundation running the V5 discovery protocol.
 func FoundationBootnodes() *Enodes {
-	nodes := &Enodes{nodes: make([]*enode.Node, len(params.MainnetBootnodes))}
-	for i, url := range params.MainnetBootnodes {
+	nodes := &Enodes{nodes: make([]*enode.Node, len(params.EtherumMainnetBootnodes))}
```
DNK90, line 74: I don't know if we can use it in the future or not. But this should be change to Ronin bootnodes instead, shouldn't it?

huyngopt1994:
> I think we should remove all ethereum network. Just add 2 types: our saigon testnet and mainnet. It would prevent misconfiguration from users

Yes let me separate in to the other PR for this stuff.

mobile/params.go
huyngopt1994: It's backward compatibilities in running testing right now, so we will keep it first.

minh-bq (approved):

PR #438, opened 2024-04-24
consortium-v2: query contract based on block hash

andicrypt:
* To tackle reorg-based attack, this PR changes contract query to using block hash.
* Minor fix: newLimit would be updated with new rule after Tripp effective

### DIFF
--- a/consensus/consortium/common/contract.go
+++ b/consensus/consortium/common/contract.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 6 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -484,10 +484,10 @@ func (c *Consortium) snapshot(chain consensus.ChainHeaderReader, number uint64,
 				log.Trace("Loaded snapshot from disk", "number", number, "hash", hash.Hex())
 				break
 			}
+			header := chain.GetHeaderByHash(hash)
 
-			// get validators set from number
 			_, _, _, contract := c.readSignerAndContract()
-			validators, err := contract.GetBlockProducers(big.NewInt(0).SetUint64(number))
+			validators, err := contract.GetBlockProducers(header.ParentHash, new(big.Int).SetUint64(number))
```
minh-bq, line 490: This does not use block hash because it does not reach the hardfork, so I think passing common.Hash{ } is enough

consensus/consortium/v2/consortium.go
```diff
@@ -806,9 +800,6 @@ func (c *Consortium) Prepare(chain consensus.ChainHeaderReader, header *types.He
 		if err != nil {
 			return err
 		}
-		// After Tripp, validator candidates are read only once at
```
minh-bq, line 809: Why is this comment block removed?

consensus/consortium/v2/snapshot.go
```diff
@@ -263,9 +265,10 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 
 				oldLimit := len(snap.validators())/2 + 1
 				var newLimit int
-				// After Tripp, list of block producers is retrieved from the
```
minh-bq, line 266: Overall, I think this change does not relates to this PR. This PR should be about query contract by using block hash, any other changes should be in a separate PR.

consensus/consortium/common/contract.go
```diff
@@ -273,33 +323,71 @@ func (c *ContractIntegrator) FinalityReward(opts *ApplyTransactOpts, votedValida
...
+	}
+	// do smart contract call
+	msgData := (hexutil.Bytes)(data)
+	to := c.chainConfig.ConsortiumV2Contracts.StakingContract
+	gas := (hexutil.Uint64)(uint64(math.MaxUint64 / 2))
+	result, err := c.ethAPI.Call(context.Background(), ethapi.TransactionArgs{
+		Gas:  &gas,
+		To:   &to,
+		Data: &msgData,
+	}, blockNr, nil, nil)
```
minh-bq, line 383: nit: you can consider but this ethAPI.Call into a function

consensus/consortium/common/contract.go
```diff
@@ -117,36 +123,80 @@ func NewContractIntegrator(config *chainParams.ChainConfig, backend bind.Contrac
...
 	}, nil
 }
 
 // GetBlockProducers retrieves block producer addresses
-func (c *ContractIntegrator) GetBlockProducers(blockNumber *big.Int) ([]common.Address, error) {
-	callOpts := bind.CallOpts{
-		BlockNumber: blockNumber,
+func (c *ContractIntegrator) GetBlockProducers(blockHash common.Hash, blockNumber *big.Int) ([]common.Address, error) {
+	blockNr := rpc.BlockNumberOrHashWithHash(blockHash, false)
+	if !c.chainConfig.IsTripp(blockNumber) {
```
minh-bq, line 144: I prefer flipping the condition here to to be same as other check. Something like
```
if c.chainConfig.IsTripp() {...}
```

minh-bq (approved):

PR #448, opened 2024-05-15
consortium-v2: add additional check in verify cascading fields

andicrypt:


consensus/consortium/v2/consortium.go
```diff
@@ -741,6 +745,8 @@ func (c *Consortium) getCheckpointValidatorsFromContract(
 					BlsPublicKey: blsPublicKey,
 					Weight:       weights[i],
 				})
+			} else {
+				log.Error("failed to get bls public key", "err", err, "candidate", candidate, "number", header.Number)
```
minh-bq, line 749: Should be log.Info
Should be "Failed"

minh-bq: Why log.Warn? There is case like in testnet, while migrating, some validators does not input their BLS key, so this leads to error in this case but there is not thing wrong with the node or contract side so I think log.Info is enough.

minh-bq (approved):

PR #453, opened 2024-05-21
IntermediateRoot: add flag for threshold to update concurrently

Francesco4203:
Divide the root updating of stateObjects into goroutines if number of stateObjects is at least the threshold


cmd/utils/flags.go
```diff
@@ -1071,6 +1071,13 @@ var (
 		Usage:    "List of mock bls public keys which are reflect 1:1 with mock.validators",
 		Category: flags.MockCategory,
 	}
+
+	ConcurrentUpdateThresholdFlag = &cli.IntFlag{
+		Name:     "concurrencyupdatethreashold",
```
minh-bq, line 1076: nit: concurrent-update-threshold
I think this flag name is too long and a little bit hard to read so adding some dashes looks better

core/state/state_object.go
```diff
@@ -319,7 +319,7 @@ func (s *stateObject) finalise(prefetch bool) {
 
 // updateTrie writes cached storage modifications into the object's storage trie.
 // It will return nil if the trie has not been loaded and no changes have been made
-func (s *stateObject) updateTrie(db Database) Trie {
+func (s *stateObject) updateTrie(db Database, concurrent bool) Trie {
```
minh-bq, line 322: I think change this variable to skipTrieUpdate is more reasonable

core/state/state_object.go
```diff
@@ -369,16 +376,46 @@ func (s *stateObject) updateTrie(db Database) Trie {
...
+		if value == s.originStorage[key] {
+			continue
+		}
+		s.originStorage[key] = value
+
+		var v []byte
+
+		if (value == common.Hash{}) {
+			s.setError(tr.TryDelete(key[:]))
+			s.db.StorageDeleted += 1
```
minh-bq, line 400: This is shared resource, not safe to update concurrently. Moreover, this is updated in updateTrie already

core/state/state_object.go
```diff
@@ -369,16 +376,46 @@ func (s *stateObject) updateTrie(db Database) Trie {
...
+		} else {
+			v, _ = rlp.EncodeToBytes(common.TrimLeftZeroes(value[:]))
+			s.setError(tr.TryUpdate(key[:], v))
+			s.db.StorageUpdated += 1
```
minh-bq, line 404: Same as above comment

core/state/statedb.go
```diff
@@ -855,11 +863,52 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 	// the account prefetcher. Instead, let's process all the storage updates
 	// first, giving the account prefeches just a few more milliseconds of time
 	// to pull useful data from disk.
+
+	// ---------------------------------------------- DEBUGGING -------------------------------------------------------------------------------------------
```
minh-bq, line 867: Please clean this up

core/state/statedb.go
```diff
@@ -855,11 +863,52 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+
+	// Get the stateObjects needed to be updated
+	updateObjs := []*stateObject{}
 	for addr := range s.stateObjectsPending {
 		if obj := s.stateObjects[addr]; !obj.deleted {
-			obj.updateRoot(s.db)
+			updateObjs = append(updateObjs, obj)
+			obj.updateTrie(s.db, true)
```
minh-bq, line 874: We should use `ConcurrentUpdateThreshold` config to determine if we need to update trie concurrently or not

core/state/statedb.go
```diff
@@ -855,11 +863,52 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
 		}
 	}
+	min := func(a, b int) int {
```
minh-bq, line 877: With the new way of calculating number of objects per routine, this function is only called once, so I think removing this function and using if for that one is better

core/state/state_object.go
```diff
@@ -332,26 +332,33 @@ func (s *stateObject) updateTrie(db Database) Trie {
 	// The snapshot storage map for the object
 	var storage map[common.Hash][]byte
 	// Insert all the pending updates into the trie
-	tr := s.getTrie(db)
+	var tr Trie = nil
```
minh-bq, line 335: nit: don't need to initialize to nil here, the default value is nil already

PR #459, opened 2024-05-23
rawdb: query contract code using new scheme only

minh-bq:
Commit 87c0ba92136a ("core, eth, les, trie: add a prefix to contract code (#21080)") on Ethereum changes the key scheme to save contract code. This makes the rawdb.ReadCode must query using both old and new scheme. However, Ronin starts after this commit, so all code is stored using new scheme. Remove the old scheme query to save 1 database read when read code.


huyngopt1994 (approved):
LGTM

PR #460, opened 2024-05-24
core/state: reduce memory allocation in IsAddressBlacklisted

minh-bq:
IsAddressBlacklisted is a hot function as it is called at least twice when processing a transaction in block to check the sender, recipient address in legacy transaction and 1 more with payer address in sponsored transaction. This PR tries to reduce memory allocation in this function by avoiding converting back and forth through big.Int but write the bytes directly to slice.

go test -bench=BenchmarkIsAddressBlacklisted -benchtime=5s

Before this PR
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/state
BenchmarkIsAddressBlacklisted-8          3112669              1805 ns/op            1008 B/op         16 allocs/op
```
After this PR
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/state
BenchmarkIsAddressBlacklisted-8          3698192              1668 ns/op             784 B/op          8 allocs/op
```

### DIFF
--- a/core/state/statedb_utils.go
+++ b/core/state/statedb_utils.go
@@ -34,7 +34,12 @@ var (
 	slotRoninValidatorMapping = map[string]uint64{
 		VALIDATORS: 6,
 	}
+	valueOne common.Hash
 )
+
+func init() {
+	valueOne.SetBytes([]byte{0x1})
+}
 
 // IsWhitelistedDeployer reads the contract storage to check if an address is allow to deploy
 func IsWhitelistedDeployerV2(statedb *StateDB, address common.Address, blockTime uint64, whiteListContract *common.Address) bool {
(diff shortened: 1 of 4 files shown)


core/state/statedb_utils.go
```diff
@@ -95,14 +95,17 @@ func IsAddressBlacklisted(statedb *StateDB, blacklistAddr *common.Address, addre
 	contract := *blacklistAddr
 	disabledSlot := slotBlacklistContractMapping[DISABLED]
 	disabled := statedb.GetState(contract, GetLocSimpleVariable(disabledSlot))
-	if disabled.Big().Cmp(big.NewInt(1)) == 0 {
+
+	var valueOne common.Hash
```
DNK90, line 99: could you make it as a constant instead of doing this?

core/state/statedb_utils.go
```diff
@@ -95,14 +100,15 @@ func IsAddressBlacklisted(statedb *StateDB, blacklistAddr *common.Address, addre
 	contract := *blacklistAddr
 	disabledSlot := slotBlacklistContractMapping[DISABLED]
 	disabled := statedb.GetState(contract, GetLocSimpleVariable(disabledSlot))
-	if disabled.Big().Cmp(big.NewInt(1)) == 0 {
+
+	if disabled == valueOne {
```
huyngopt1994, line 104: I think we wanna change value to variable, then it should show its purpose. `disabledStateValue` using this name is better

huyngopt1994:
seem a little cons related to timestamp for full ops 
6168584256 (After)
5618367545 (Before)

huyngopt1994 (approved):

PR #461, opened 2024-05-27
core/vm: change Consortium precompiled mapping into global variable

minh-bq:
precompile query is a hot function which is called in every call related opcode.
This commit tries to optimize this function by making the Consortium precompiled
contract mapping into global variable so that precompile query does not need to
dynamically allocate the mapping every time it is called.

~ go test -bench=BenchmarkConsortiumPrecompileQuery -benchtime=5s -tags=blst_enabled

Before:
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/vm
BenchmarkConsortiumPrecompileQuery-8     4216357              2402 ns/op            1166 B/op          9 allocs/op
```
After:
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/vm
BenchmarkConsortiumPrecompileQuery-8    100000000               50.40 ns/op           24 B/op          1 allocs/op
```

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -20,6 +20,8 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"errors"
+	"math/big"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -27,7 +29,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto/bls12381"
 	"github.com/ethereum/go-ethereum/crypto/bn256"
 	"github.com/ethereum/go-ethereum/params"
-	"math/big"
 
 	//lint:ignore SA1019 Needed for precompile
 	"golang.org/x/crypto/ripemd160"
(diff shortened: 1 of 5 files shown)


core/vm/contracts.go
```diff
@@ -106,13 +57,76 @@ var PrecompiledContractsBLS = map[common.Address]PrecompiledContract{
...
+	// PrecompiledContractsBerlin contains the default set of pre-compiled Ethereum
+	// contracts used in the Berlin release.
+	PrecompiledContractsBerlin map[common.Address]PrecompiledContract
 )
 
+func copyPrecompiledContract(input map[common.Address]PrecompiledContract) map[common.Address]PrecompiledContract {
+	output := make(map[common.Address]PrecompiledContract)
+
+	for address, contract := range input {
+		output[address] = contract
```
huyngopt1994, line 101: naming, maybe 
```
input: src 
output: dst
```

core/vm/evm_test.go
```diff
@@ -740,3 +740,72 @@ func TestInternalTransactionOrderDelegateCalls(t *testing.T) {
...
+	if p.caller != caller {
+		t.Fatal("Incorrect caller in precompiled contract")
+	}
+
+	precompiledContract, ok = evm.precompile(caller, common.BytesToAddress([]byte{1}))
+	if !ok {
+		t.Fatal("Failed to get ecrecover precompiled contract")
+	}
+	if _, ok := precompiledContract.(*ecrecover); !ok {
+		t.Fatal("Wrong ecreocver contract")
```
huyngopt1994, line 775: ecrecover

minh-bq: fixed

core/vm/contracts.go
minh-bq: I've changed to different name. I don't like src and dst as it feels like a move operation.

andicrypt (approved):

PR #462, opened 2024-05-29
consortium-v2/state: upgrade Axie proxy contract

andicrypt:
**Aaron** hardfork draft: 
- [x] Upgrade Axie and Land proxy code to achieve transparent proxy.   

- [x] Introduce block producer bit set in header's extra to replace the list of block producers.

   Block producer set is a subset of validator candidate set, but currently, both set are stored in plain, which costs unnecessary resource. This commit introduces BlockProducersBitSet (uint64) as an encoded representation of block producer list, indexing to validator candidate list. The original block producer set can be reconstructed from the bit set and validator candidate set. As a result, instead of storing a list of block producers in header extra, only a single bit set of type uint64 is stored.


huyngopt1994:
@andicrypt Could we put unitest here ?

andicrypt:
> @andicrypt Could we put unitest here ?

Do you mean a new Ronin-test suties for these changes? I just added some unit test for these changes.

params/config.go
```diff
@@ -575,13 +578,19 @@ type ChainConfig struct {
...
 }
 
 type ContractUpgrade struct {
 	ProxyAddress          common.Address `json:"proxyAddress"`
 	ImplementationAddress common.Address `json:"implementationAddress"`
 }
 
+type ContractCodeUpgrade struct {
+	CodeAddress common.Address `json:"contractAddress"`
+	Code        []byte         `json:"code"`
```
minh-bq, line 591: https://pkg.go.dev/encoding/json#Marshal

Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

So in the genesis.json, the code field must be base64 of real code. Is this an expected behavior?

consensus/consortium/v2/consortium_test.go
```diff
@@ -1825,6 +1825,9 @@ func TestUpgradeRoninTrustedOrg(t *testing.T) {
 		Consortium: &params.ConsortiumConfig{
 			EpochV2: 200,
 		},
+		ConsortiumV2Contracts: &params.ConsortiumV2Contracts{
```
minh-bq, line 1828: Why changing this?

params/config.go
andicrypt: You are correct. I have changed the type of Code to string (hex), which will be loaded from genesis file, before being decoded into []byte for statedb process.

consensus/consortium/v2/consortium.go
```diff
@@ -837,11 +848,23 @@ func (c *Consortium) Prepare(chain consensus.ChainHeaderReader, header *types.He
...
 		if c.IsTrippEffective(chain, header) {
 			if c.IsPeriodBlock(chain, header) {
-				extraData.CheckpointValidators = checkpointValidator
+				extraData.CheckpointValidators = checkpointValidators
+			} else {
+				// checkpointValidators get from contract is nil at epoch block; hence,
+				// checkpointValidators is reassigned by the candidate list from snapshot,
+				// which is used in encodingValidatorBitSet step.
+				checkpointValidators = snap.ValidatorsWithBlsPub
+			}
```
minh-bq, line 857: I've just realized this is incorrect, we must always use snap.ValidatorsWithBlsPub. At period block, we don't switch to new block producer set right way but wait until len(blockProducer) / 2 (logic in snapshot). So the block producer here must be the indices of block producers in the old set in snap.ValidatorsWithBlsPub not the new set in extradata.

minh-bq: This is an incorrect comment

minh-bq (approved):

PR #463, opened 2024-06-05
consortium-v2: ensure checkpointValidators is empty in non-period block

minh-bq:
Ensure checkpointValidators field in header's extradata is empty in non-period
block.

### DIFF
--- a/consensus/consortium/v2/consortium_test.go
+++ b/consensus/consortium/v2/consortium_test.go
@@ -1083,6 +1083,7 @@ func TestGetCheckpointValidatorFromContract(t *testing.T) {
 			EpochV2: 200,
 		},
 		contract:           mock,
+		isTest:             true,
 		testTrippEffective: true,
 		testTrippPeriod:    false,
 	}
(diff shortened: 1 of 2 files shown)


consensus/consortium/v2/consortium_test.go
```diff
@@ -2269,3 +2270,53 @@ func TestIsTrippEffective(t *testing.T) {
...
+			{},
+		},
+		BlockProducers: []common.Address{
+			{},
+		},
+	}
+	err = c.verifyValidatorFieldsInExtraData(nil, &extraData, &header)
+	if err == nil {
+		t.Fatalf("Expect an error")
+	}
```
khanhtaskymaviscom, line 2321: I think we should add test for the period block as well. I don't see any other test cover the case.

minh-bq: This test focuses on must be empty cases, the must have check (e.g. at period block, len(checkpointValidators) must not be 0) is redundant I think, as there are checks at Finalize. We may clean up these must have checks after this release.

khanhtaskymaviscom (approved):

andicrypt (approved):

PR #469, opened 2024-06-12
core/vm: only log when the opcode is registered

minh-bq:
We move the log and transaction hash calculation inside the if, so that they are executed only when the opcode is registered in publish event.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -669,13 +669,12 @@ func (evm *EVM) PublishEvent(
 ) {
 	context := evm.Context
 	if context.CurrentTransaction == nil {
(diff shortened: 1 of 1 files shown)


core/vm/evm.go
```diff
@@ -669,13 +669,12 @@ func (evm *EVM) PublishEvent(
 ) {
 	context := evm.Context
 	if context.CurrentTransaction == nil {
-		log.Debug("[EVM] PublishEvent - Transaction is nil", "height", context.BlockNumber.Int64())
```
huyngopt1994, line 672: I think we still could keep this debug log for tracking logic currentTransaction is nil.

PR #470, opened 2024-06-13
consortium-v2: fix wrong parenthesis placement in non-epoch block check

minh-bq:
Fix wrong parenthesis placement in non-epoch block check in verifyValidatorFieldsInExtraData which leads to incorrect check result.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -362,32 +362,65 @@ func (c *Consortium) verifyValidatorFieldsInExtraData(
 ) error {
 	isEpoch := header.Number.Uint64()%c.config.EpochV2 == 0 || c.chainConfig.IsOnConsortiumV2(header.Number)
 	if !isEpoch && (len(extraData.CheckpointValidators) != 0 || len(extraData.BlockProducers) != 0) || extraData.BlockProducersBitSet != 0 {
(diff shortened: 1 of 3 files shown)


consensus/consortium/v2/consortium.go
```diff
@@ -361,7 +361,7 @@ func (c *Consortium) verifyValidatorFieldsInExtraData(
 	header *types.Header,
 ) error {
 	isEpoch := header.Number.Uint64()%c.config.EpochV2 == 0 || c.chainConfig.IsOnConsortiumV2(header.Number)
-	if !isEpoch && (len(extraData.CheckpointValidators) != 0 || len(extraData.BlockProducers) != 0) || extraData.BlockProducersBitSet != 0 {
+	if !isEpoch && (len(extraData.CheckpointValidators) != 0 || len(extraData.BlockProducers) != 0 || extraData.BlockProducersBitSet != 0) {
```
huyngopt1994, line 364: could we move logic into `(len(extraData.CheckpointValidators) != 0 || len(extraData.BlockProducers) != 0 || extraData.BlockProducersBitSet != 0)` variable then use it it in if clause ?

minh-bq: changed

huyngopt1994 (approved):

andicrypt (approved):

PR #472, opened 2024-06-17
core/vm: add EVM benchmark tests

minh-bq:
This PR aim to add real world contracts to our EVM benchmark


.github/workflows/evm-benchmark.yml
```diff
@@ -0,0 +1,52 @@
...
+          fetch-depth: 0
+
+      - name: 'Setup Go'
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 #v3.3.0
+        with:
+          go-version: '1.20.10'
+
+      - name: 'Run EVM benchmark'
+        run: |
+          go install golang.org/x/perf/cmd/benchstat@latest
```
khangvv, line 44: we should pin `benchstat` version instead of using latest

minh-bq: fixed

.github/workflows/evm-benchmark.yml
```diff
@@ -0,0 +1,52 @@
+name: EVM benchmark
```
huyngopt1994, line 1: @minh-bq we can add some infor in the name here for easily tracking 

For example 
```
run-name: E2E Reuse Test ${{ inputs.rpc_endpoint }}, path ${{ inputs.test_path }} by ${{ github.actor }}
```

.github/workflows/evm-benchmark.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 #v3.3.0
+        with:
+          go-version: '1.20.10'
+
+      - name: 'Run EVM benchmark'
+        run: |
+          go install golang.org/x/perf/cmd/benchstat@v0.0.0-20240604174448-3b48cf0e0164
+          cd core/vm
+          if [[ ${{ github.event_name }} = "pull_request" ]]; then
+            ../../script/benchstat.sh -c "go test -test.v -run=^$ -bench=BenchmarkEvm -count=10" \
```
huyngopt1994, line 47: -count=10 make it input too ? maybe set default is 10

.github/workflows/evm-benchmark.yml
```diff
@@ -0,0 +1,52 @@
+name: EVM benchmark
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - core/vm/**
+  workflow_dispatch:
```
huyngopt1994, line 9: Not sure we can run 2 modes in same file or not. If not,  We need to separate 2 files, one for
```
on:
  pull_request:
    branches:
      - master
    paths:
      - core/vm/**
 ```

one for workflow_dispatch
```
on:
  workflow_dispatch:
```

then remove if logic.

File: .github/workflows/evm-benchmark.yml (hunk 3 above)
Comment (minh-bq) [reply .github/workflows/evm-benchmark.yml:47]: This tool requires the benchmark to run at least 10 times to make the benchmark statistically significant

File: .github/workflows/evm-benchmark.yml (hunk 2 above)
Comment (minh-bq) [reply .github/workflows/evm-benchmark.yml:1]: What information that you think is needed here?

File: .github/workflows/evm-benchmark.yml (hunk 3 above)
Comment (huyngopt1994) [reply .github/workflows/evm-benchmark.yml:47]: ya I means, we can put it like input (optional) with default is 10, then defautly it still use 10, but we can adjust in case we need.

Comment (huyngopt1994) [reply .github/workflows/evm-benchmark.yml:47]: like this

```
      go_version:
        description: 'The Go version to download (if necessary) and use.'
        required: false
        type: string
        default: '1.21.3'
```

.github/workflows/evm-benchmark.yml
huyngopt1994: old commit - new commit - author
For quickly checking if we triggering manual mode.

.github/workflows/evm-benchmark.yml
minh-bq: I just feel this is unnecessary, AFAIK, there is no scenario that we need to change that number.

PR #475, opened 2024-06-18
core/vm: merge opJumpdest to opJump/opJumpi

Francesco4203:
Make opJumpdest to be executed right after opJump/opJumpi:
- Reduce time of unnecessary checking for opJumpdest properties
- Reduce overhead of fail branch predicting from opJump/opJumpi to opJumpdest (since it is always opJumpdest after)

Benchmark:
./benchstat.sh -c "go test -test.v -run=^$ -bench=BenchmarkRBTreeContract -tags=blst_enabled -benchtime=100x -count=10" -o HEAD~1 -n HEAD

Result:
```
goos: darwin
goarch: arm64
pkg: github.com/ethereum/go-ethereum/core/vm
                  │   old.txt   │              new.txt               │
                  │   sec/op    │   sec/op     vs base               │
RBTreeContract-12   109.2m ± 2%   104.2m ± 1%  -4.56% (p=0.000 n=10)
```


core/vm/evm_test.go
```diff
@@ -740,3 +743,115 @@ func TestInternalTransactionOrderDelegateCalls(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkRBTreeContract(b *testing.B) {
```
minh-bq, line 747: Can you move and adapt this test case to our new evm_benchmark_test structure?

PR #480, opened 2024-06-24
p2p/discover: apply enr filtering in discovery phase, fix minor logic in doRevalidate

huyngopt1994:
**Description**  

Node discovery filtering based on ENR records for Ronin Network

Current implementation of discovery protocol is chain agnostic and does not differentiate between different chains (ETH, Ronin, ...). The discover protocol currently gossips peers from networks with different chain ids. This causes a pretty significant slowdown during peer discovery.

- [x] Apply ENR filter Metrics in the discovery phase
- [x] Build ENR filter success result metrics  

Reference by  https://github.com/bnb-chain/bsc/pull/1320

Dial and Enr Filter ratio metrics for an RPC node in the recent 3 hour metrics 

Before 

Screenshot 2024-06-24 at 4 14 24 PM

Screenshot 2024-06-24 at 4 17 57 PM

After 
Screenshot 2024-06-24 at 4 14 14 PM

Screenshot 2024-06-24 at 4 18 06 PM

The filtering ratio is ~9x now. 

For the Dial Result 
The formula is 

```
sum(rate(p2p_dials_success{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_forkid{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_genesis{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_network{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_peer{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_timeout{statefulset_kubernetes_io_pod_name=~"$pod"}[1m]) - rate(eth_protocols_eth_egress_handshake_error_version{statefulset_kubernetes_io_pod_name=~"$pod"}[1m])) by (statefulset_kubernetes_io_pod_name)  / sum(rate(p2p_dials{statefulset_kubernetes_io_pod_name=~"$pod"}[1m])) by (statefulset_kubernetes_io_pod_name)
```
It will show the dial success ratio based on 1 m interval.

For the Dial Result (1h)

It's the summarization of the total of dialing including failed/passed error codes. Overall the number of calling dial and dial results seem to improve after applying. Let us monitor that node for more days to get more metrics.

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -43,7 +43,7 @@ const (
 	alpha                  = 3               // Kademlia concurrency factor
 	bucketSize             = 16              // Kademlia bucket size
 	maxReplacements        = 10              // Size of per-bucket replacement list
-	maxWorkerTask          = 60              // Maximum number of worker tasks
+	maxWorkerTask          = 90              // Maximum number of worker tasks
 	timeoutWorkerTaskClose = 1 * time.Second // Timeout for waiting workerPoolTask is refill full
 	// We keep buckets for the upper 1/15 of distances because
 	// it's very unlikely we'll ever encounter a node that's closer.
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -40,10 +40,11 @@ import (
 )
 
 const (
-	alpha           = 3   // Kademlia concurrency factor
-	bucketSize      = 16  // Kademlia bucket size
-	maxReplacements = 10  // Size of per-bucket replacement list
-	maxWorkerTask   = 100 // Maximum number of worker tasks
+	alpha                  = 3               // Kademlia concurrency factor
+	bucketSize             = 16              // Kademlia bucket size
+	maxReplacements        = 10              // Size of per-bucket replacement list
+	maxWorkerTask          = 60              // Maximum number of worker tasks
+	timeoutWorkerTaskClose = 1 * time.Second // Timeout for waiting workerPoolTask is refill full
 	// We keep buckets for the upper 1/15 of distances because
 	// it's very unlikely we'll ever encounter a node that's closer.
 	hashBits          = len(common.Hash{}) * 8
@@ -203,6 +204,25 @@ func (tab *Table) getNode(id enode.ID) *enode.Node {
 	return nil
 }
 
+func (tab *Table) closeWorkerTask() {
+	waitTicker := time.NewTicker(1 * time.Millisecond)
+	defer waitTicker.Stop()
+	timeoutChan := time.After(timeoutWorkerTaskClose)
+	for {
+		select {
+		case <-waitTicker.C:
+			if len(tab.workerPoolTask) == cap(tab.workerPoolTask) {
+				log.Info("workerPoolTask is refill full, closing channel.")
+				return
+			}
+
+		case <-timeoutChan:
+			log.Warn("Timeout waiting for workerPoolTask is refill full , force exit it.")
+			return
+		}
+	}
+}
+
 // close terminates the network listener and flushes the node database.
 func (tab *Table) close() {
 	close(tab.closeReq)
@@ -301,6 +321,8 @@ loop:
 	if revalidateDone != nil {
 		<-revalidateDone
 	}
+
+	tab.closeWorkerTask()
 	close(tab.closed)
 }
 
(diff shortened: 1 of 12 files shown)


p2p/discover/table.go
```diff
@@ -481,10 +489,18 @@ func (tab *Table) bucketAtDistance(d int) *bucket {
 //
 // The caller must not hold tab.mutex.
 func (tab *Table) addSeenNode(n *node) {
+	go tab.addSeenNodeSync(n)
```
minh-bq, line 492: This can be a DOS vector since we just spawn new goroutines without max threshold

p2p/discover/table.go
```diff
@@ -347,10 +349,16 @@ func (tab *Table) doRevalidate(done chan<- struct{}) {
 
 	// Also fetch record if the node replied and returned a higher sequence number.
 	if last.Seq() < remoteSeq {
-		n, err := tab.net.RequestENR(unwrapNode(last))
-		if err != nil {
-			tab.log.Debug("ENR request failed", "id", last.ID(), "addr", last.addr(), "err", err)
+		n, enrErr := tab.net.RequestENR(unwrapNode(last))
```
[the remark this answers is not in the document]
huyngopt1994: I think it should not, the previous logic will not add this node into bucket when it's failed for requestingENR

huyngopt1994:
@minh-bq Pushed commits for handling your concerns, please check.

p2p/discover/table.go
```diff
@@ -494,10 +500,23 @@ func (tab *Table) bucketAtDistance(d int) *bucket {
 //
 // The caller must not hold tab.mutex.
 func (tab *Table) addSeenNode(n *node) {
-	go tab.addSeenNodeSync(n)
+	select {
+	case <-tab.workerPoolTask:
+		go tab.addSeenNodeSync(n)
```
minh-bq, line 505: I'm not so sure about this but when shutting down the node, table.close() is called and it seems like these goroutines are created does not sync with the shutdown event. I'm not sure if it safe, as I see some other old code does not sync with shutdown event either.

p2p/discover/table.go
```diff
@@ -40,10 +40,10 @@ import (
...
 
 const (
-	alpha           = 3  // Kademlia concurrency factor
-	bucketSize      = 16 // Kademlia bucket size
-	maxReplacements = 10 // Size of per-bucket replacement list
-
+	alpha           = 3   // Kademlia concurrency factor
+	bucketSize      = 16  // Kademlia bucket size
+	maxReplacements = 10  // Size of per-bucket replacement list
+	maxWorkerTask   = 100 // Maximum number of worker tasks
```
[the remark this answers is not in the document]
huyngopt1994: It's shared a pool for addVerified and addSeenNode too, 100 goroutines may not be a big deal in the worst case, Setting a low value maybe can make the performance of peering decrease so much.

p2p/discover/table.go
huyngopt1994: I don't think we need to close the channel here, as documentation, the channel is cleaned up by GC, and force closing the channel in the close method maybe needed in some cases to terminate a loop range. For our cases, we don't have a loop here.

minh-bq: I mean the concern is whether the code in goroutine can be safely executed when the table is closed

p2p/discover/table.go
```diff
@@ -341,24 +349,35 @@ func (tab *Table) doRevalidate(done chan<- struct{}) {
 		// No non-empty bucket found.
 		return
 	}
-
+	var errHandle error
 	// Ping the selected node and wait for a pong.
 	remoteSeq, err := tab.net.ping(unwrapNode(last))
 
+	if err != nil {
```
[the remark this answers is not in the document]
huyngopt1994: If the err!=nil, the remoteSeq is returned 0, and the  `if last.Seq() < remoteSeq` always is false here. Seem the old code is a little hard to read, but the logic is kept here.

minh-bq (approved):
LGTM.
Need to change the PR title to match the current convention when merging.

PR #485, opened 2024-07-08
consortium-v2: add parent list parameter to IsPeriodBlock

minh-bq:
We get this warning and error log
```
WARN [07-08|04:18:15.866] Fail to get snapshot at                  blockNumber=36,192,799 blockHash=c19608..2c657b err="unknown ancestor"
ERROR[07-08|04:18:15.866] Failed to get block                      number=36,192,799 hash=0xc19608d700837c8ab4c184791bc3522766a48efb5f2e7734b827b3f9382c657b
```
IsPeriodBlock can be called in verifyHeader path. In that path, the parents of currently verifying block may not be inserted into chain yet so querying the parent may fail. In some functions like snapshot which IsPeriodBlock uses have the parent list parameter to query parent blocks when they are not inserted into chain. This commit adds a new parent list parameter to IsPeriodBlock and passes the parent list when it is called in verifyHeader path.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -359,6 +359,7 @@ func (c *Consortium) verifyValidatorFieldsInExtraData(
 	chain consensus.ChainHeaderReader,
 	extraData *finality.HeaderExtraData,
 	header *types.Header,
+	parents []*types.Header,
 ) error {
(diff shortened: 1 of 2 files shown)


andicrypt:
LGTM

andicrypt (approved):

PR #491, opened 2024-07-11
core/types: add blob transaction type

andicrypt:
This PR introduces the blob transaction types as defined in EIP-4844, extending the generic transaction interface. A new signer is included, which incorporates BlobGasFeeCap and BlobHashes into the transaction digest. Additionally, the rlp module has been enhanced with new methods, BytesSize and StringSize, to compute the size of the encoded data. These methods are essential for calculating the RLP encoded data size of blob transactions.

### DIFF
--- a/core/types/access_list_tx.go
+++ b/core/types/access_list_tx.go
@@ -107,6 +107,10 @@ func (tx *AccessListTx) nonce() uint64          { return tx.Nonce }
 func (tx *AccessListTx) to() *common.Address    { return tx.To }
 func (tx *AccessListTx) expiredTime() uint64    { return 0 }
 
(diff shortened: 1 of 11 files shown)


minh-bq (commented):
I think
- core: improve validation loop in InsertReceiptChain
- core: check absence of blob sidecar in InsertReceiptChain
should be in different commits where we fix up the import chain

Why do we need core: fix review comment commit?

minh-bq:
Can we split the commit core/types,rlp: add blob tx without sidecar and unit test and make the related changes
```
- To *common.Address // `rlp:"nil"` // nil means contract creation
+ To common.Address // In EIP4844, to value must be non-nil
```
into a separated commit?

minh-bq (approved):

PR #492, opened 2024-07-17
consensus: add consensus blob check

Francesco4203:
Verify the new blob's fields in block's header, including BlobGasUsed, ExcessBlobGas, BlobCommitment.


consensus/consortium/v2/consortium.go
```diff
@@ -203,6 +207,49 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
 	return header.Coinbase, nil
 }
 
+// VerifyBlobHeader verifies the header of a blob block
+func (c *Consortium) VerifyBlobHeader(block *types.Block, blobs *[]kzg4844.Blob, proofs *[]kzg4844.Proof) error {
```
minh-bq, line 211: Should write some unit test for these functions

minh-bq, line 211: blobs and proofs can be passed as slices instead of pointers to slice

consensus/consortium/v2/consortium.go
minh-bq, line 1: 3 new fields in header must be included into hash function `encodeSigHeader`

consensus/consortium/v2/consortium.go
```diff
@@ -208,18 +208,18 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
...
+func (c *Consortium) VerifyBlobHeader(block *types.Block, blobs []kzg4844.Blob, proofs []kzg4844.Proof) error {
 	if err := c.verifyVersionHash(block); err != nil {
 		return err
 	}
 	if c.skipBlobCheck(block.Header()) {
 		return nil
 	}
-	if len(block.Header().BlobCommitments) != len(*blobs) || len(block.Header().BlobCommitments) != len(*proofs) {
-		return fmt.Errorf("mismatching number of blobs, commitments, and proofs. blobs: %d, commitments: %d, proofs: %d", len(*blobs), len(block.Header().BlobCommitments), len(*proofs))
+	if len(block.Header().BlobCommitments) != len(blobs) || len(block.Header().BlobCommitments) != len(proofs) {
```
minh-bq, line 218: We should create a variable `header := block.Header()` because block.Header will create a new copied header every time it is called

consensus/consortium/v2/consortium.go
```diff
@@ -208,18 +208,18 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
...
+		return fmt.Errorf("mismatching number of blobs, commitments, and proofs. blobs: %d, commitments: %d, proofs: %d", len(blobs), len(block.Header().BlobCommitments), len(proofs))
 	}
-	for i := range *blobs {
-		if err := kzg4844.VerifyBlobProof(&(*blobs)[i], block.Header().BlobCommitments[i], (*proofs)[i]); err != nil {
+	for i := range blobs {
+		if err := kzg4844.VerifyBlobProof(&(blobs)[i], block.Header().BlobCommitments[i], (proofs)[i]); err != nil {
```
minh-bq, line 222: nit: &(blobs)[i] -> &blobs[i], (proofs)[i] -> proofs[i]

consensus/consortium/v2/consortium.go
```diff
@@ -555,6 +553,10 @@ func (c *Consortium) verifyCascadingFields(chain consensus.ChainHeaderReader, he
 		if err := misc.VerifyEIP4844Header(parent, header); err != nil {
 			return err
 		}
+	} else {
+		if header.BlobCommitments != nil || header.BlobGasUsed != nil || header.ExcessBlobGas != nil {
```
minh-bq, line 557: With blob commitments, I think we should check len(header.BlobCommitments)  != 0

consensus/consortium/v2/consortium.go
```diff
@@ -1833,6 +1835,15 @@ func encodeSigHeader(w io.Writer, header *types.Header, chainId *big.Int) {
 	if header.BaseFee != nil {
 		enc = append(enc, header.BaseFee)
 	}
+	if header.BlobGasUsed != nil {
+		enc = append(enc, header.BlobGasUsed)
+	}
+	if header.BlobCommitments != nil {
```
minh-bq, line 1841: Same as the above comment, should check the length

consensus/consortium/v2/consortium.go
minh-bq: I don't test it yet but looking at the code: https://github.com/axieinfinity/ronin/blob/6feb34d9f01183f5b635868816aecad6a0526b37/rlp/decode.go#L310-L313
When rlp decoding, if the field is a slice and empty, it set the slice to 0-length slice not a nil. So I think check length is safer here

consensus/consortium/v2/consortium_test.go
```diff
@@ -2532,3 +2536,77 @@ func TestEncodeDecodeValidatorBitSet(t *testing.T) {
...
+func randBlob() *kzg4844.Blob {
+	var blob kzg4844.Blob
+	for i := 0; i < len(blob); i += gokzg4844.SerializedScalarSize {
+		fieldElementBytes := randFieldElement()
+		copy(blob[i:i+gokzg4844.SerializedScalarSize], fieldElementBytes[:])
+	}
+	return &blob
+}
+
+func TestVerifyBlobHeader(t *testing.T) {
```
minh-bq, line 2561: I think this test does not have a good coverage. skipBlobCheck will return true and we don't reach the below checks. In unit test for this case, you should create an invalid version hash, check if it fails as expected. Then correct the version hash, check both 2 cases that skipBlobCheck returns true and false. And continue doing the same for the following checks.

consensus/consortium/v2/consortium.go
Francesco4203: fixed

consensus/consortium/v2/consortium.go
Francesco4203: fixed

consensus/consortium/v2/consortium.go
Francesco4203: I think you are right, fixed

consensus/consortium/v2/consortium_test.go
Francesco4203: I have added more tests to cover all cases

consensus/consortium/v2/consortium.go
```diff
@@ -203,6 +207,51 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
...
+	}
+	for i := range blobs {
+		if err := kzg4844.VerifyBlobProof(&blobs[i], commitments[i], proofs[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Consortium) verifyVersionHash(block *types.Block, commitments []kzg4844.Commitment) error {
```
andicrypt, line 230: add comment for this function

consensus/consortium/v2/consortium_test.go
```diff
@@ -2532,3 +2536,174 @@ func TestEncodeDecodeValidatorBitSet(t *testing.T) {
...
+	// Create 100 random txs
+	var txs []*types.Transaction
+	var blobs []kzg4844.Blob
+	var commitments []kzg4844.Commitment
+	var proofs []kzg4844.Proof
+
+	var fakeCommitments []kzg4844.Commitment
+	var fakeProofs []kzg4844.Proof
+
+	for i := 0; i < 100; i++ {
```
andicrypt, line 2580: I think you can move these logic to a seperate func, which produces a mock block with blob tx. We may need to use it in later test.

consensus/consortium/v2/consortium_test.go
```diff
@@ -2532,3 +2536,174 @@ func TestEncodeDecodeValidatorBitSet(t *testing.T) {
...
+		nBlobs, err := rand.Int(rand.Reader, big.NewInt(7))
+		if err != nil {
+			t.Fatal(err)
+		}
+		hasher := sha256.New()
+		var blobHashes []common.Hash
+		for j := 0; j < int(nBlobs.Int64()); j++ {
```
andicrypt, line 2587: Similarly, a separate func to create mock blob tx

consensus/consortium/v2/consortium.go
```diff
@@ -203,6 +207,51 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
 	return header.Coinbase, nil
 }
 
+// VerifyBlobHeader verifies the header of a blob block
+func (c *Consortium) VerifyBlobHeader(block *types.Block, blobs []kzg4844.Blob, proofs []kzg4844.Proof, commitments []kzg4844.Commitment) error {
```
andicrypt, line 211: I think beside these checks, we need to ensure that the number of blobs per block must not larger than `MaxBlobsPerBlock = 6` you can add this config to `protocol_params.go` file

consensus/consortium/v2/consortium_test.go
```diff
@@ -2532,3 +2536,174 @@ func TestEncodeDecodeValidatorBitSet(t *testing.T) {
...
+	var txs []*types.Transaction
+	var blobs []kzg4844.Blob
+	var commitments []kzg4844.Commitment
+	var proofs []kzg4844.Proof
+
+	var fakeCommitments []kzg4844.Commitment
+	var fakeProofs []kzg4844.Proof
+
+	for i := 0; i < 100; i++ {
+		nBlobs, err := rand.Int(rand.Reader, big.NewInt(7))
```
andicrypt, line 2581: Lastly, as we use this rand number, we may not sure the number of blobs produced, so I recommend using more deterministic logic, which allows us to verify the case of `number of blobs per block` < `MaxBlobsPerBlock`.

consensus/consortium/main.go
```diff
@@ -48,6 +49,15 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
 	return c.v1.Author(header)
 }
 
+// VerifyBlobHeader implements consensus.Engine, verifying the header's blob commitment
+func (c *Consortium) VerifyBlobHeader(block *types.Block, blobs []kzg4844.Blob, proofs []kzg4844.Proof, commitments []kzg4844.Commitment) error {
+	if c.chainConfig.IsConsortiumV2(block.Header().Number) {
```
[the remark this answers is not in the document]
Francesco4203: fixed

consensus/consortium/v2/consortium.go
Francesco4203: added

consensus/consortium/v2/consortium_test.go
Francesco4203: I have moved them to separate functions

consensus/consortium/v2/consortium.go
Francesco4203: You are right I forgot to check that, added

consensus/consortium/v2/consortium_test.go
Francesco4203: fixed

consensus/consortium/v2/consortium.go
```diff
@@ -203,6 +209,89 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
...
+		if err := c.verifySidecar(sidecar); err != nil {
+			return err, nil
+		}
+	}
+	blobSidecars := types.BlobSidecars{}
+	curIndex := 0
+	for _, tx := range block.Transactions() {
+		if tx.Type() == types.BlobTxType {
+			blobSidecars = append(blobSidecars, &types.BlobSidecar{
+				BlobTxSidecar: sidecars[curIndex],
```
minh-bq, line 238: need curIndex++ in this if or just use tx.BlobTxSidecar()

Francesco4203: Oh I forgot to use tx.BlobTxSidecar(), fixed

Francesco4203: Well actually, there is no sidecars in the block here (and therefore no sidecars in the tx itself), I guess we will just add curIndex++

PR #505, opened 2024-07-29
eth: add eth/68 and eth/100

Francesco4203:
This PR is to:
- Cherry-pick ETH68, which added `Types` and `Sizes` to `TransactionHashesPacket`. `Sizes` will be useful for blob transactions, which are large in size, as it helps prevent bandwidth overload in the network.
- Implement ETH100, adding new blob's field `Sidecars` to `BlockBodiesPacket` when broadcasting and receiving new blocks.

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -230,7 +230,7 @@ func (bc *BlockChain) GetBlobSidecarsByNumber(number uint64) types.BlobSidecars 
 	if sidecars, ok := bc.blobSidecarsCache.Get(hash); ok {
 		return sidecars.(types.BlobSidecars)
 	}
-	
+
 	sidecars := rawdb.ReadBlobSidecars(bc.db, hash, number)
 	bc.blobSidecarsCache.Add(hash, sidecars)
 	return sidecars
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -549,10 +549,10 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 		// Send the block to a subset of our peers
 		transfer := peers[:int(math.Sqrt(float64(len(peers))))]
 		for _, peer := range transfer {
-			var sidecars []types.BlobTxSidecar
+			var sidecars []*types.BlobTxSidecar
 			if peer.Version() >= eth.ETH100 {
 				for _, blobSidecar := range rawdb.ReadBlobSidecars(h.database, block.Hash(), block.NumberU64()) {
-					sidecars = append(sidecars, (*blobSidecar).BlobTxSidecar)
+					sidecars = append(sidecars, &blobSidecar.BlobTxSidecar)
 				}
 			}
 			peer.AsyncSendNewBlock(block, td, sidecars)
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/forkid"
+	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vote"
@@ -548,7 +549,13 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 		// Send the block to a subset of our peers
 		transfer := peers[:int(math.Sqrt(float64(len(peers))))]
 		for _, peer := range transfer {
-			peer.AsyncSendNewBlock(block, td)
+			var sidecars []types.BlobTxSidecar
+			if peer.Version() >= eth.ETH100 {
+				for _, blobSidecar := range rawdb.ReadBlobSidecars(h.database, block.Hash(), block.NumberU64()) {
+					sidecars = append(sidecars, (*blobSidecar).BlobTxSidecar)
+				}
+			}
+			peer.AsyncSendNewBlock(block, td, sidecars)
 		}
 		log.Trace("Propagated block", "hash", hash, "recipients", len(transfer), "duration", common.PrettyDuration(time.Since(block.ReceivedAt)))
 		return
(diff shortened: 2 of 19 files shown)


eth/protocols/eth/protocol.go
```diff
@@ -39,11 +42,11 @@ const ProtocolName = "eth"
 
 // ProtocolVersions are the supported versions of the `eth` protocol (first
 // is primary).
-var ProtocolVersions = []uint{ETH66}
+var ProtocolVersions = []uint{ETH100, ETH68, ETH67, ETH66}
```
minh-bq, line 45: This should be ETH100, ETH66 only. We will transition from ETH66 to ETH100 directly.

eth/protocols/eth/protocol.go
```diff
@@ -212,6 +235,12 @@ type BlockBodiesPacket66 struct {
 	BlockBodiesPacket
 }
 
+// BlockBodiesPacket100 is the network packet for block content distribution over eth/100.
+type BlockBodiesPacket100 struct {
+	RequestId   uint64
+	BlockBodies []BlockBody100
```
minh-bq, line 241: I think we should use []*BlockBody100, which is potentially faster when passing/copying this struct around.

eth/protocols/eth/protocol.go
```diff
@@ -229,6 +258,13 @@ type BlockBody struct {
 	Uncles       []*types.Header      // Uncles contained within a block
 }
 
+// BlockBody100 represents the data content of a single block over eth/100.
+type BlockBody100 struct {
+	Transactions []*types.Transaction  // Transactions contained within a block
+	Uncles       []*types.Header       // Uncles contained within a block
+	Sidecars     []types.BlobTxSidecar // Sidecar data for the block
```
minh-bq, line 265: I also prefer []*types.BlobTxSidecar here

eth/protocols/eth/protocol.go
```diff
@@ -242,6 +278,20 @@ func (p *BlockBodiesPacket) Unpack() ([][]*types.Transaction, [][]*types.Header)
 	return txset, uncleset
 }
 
+// Unpack retrieves the transactions, uncles, blobs, proofs, and commitments from the range packet and returns
+// them in a split flat format that's more consistent with the internal data structures.
+func (p *BlockBodiesPacket100) Unpack() ([][]*types.Transaction, [][]*types.Header, [][]types.BlobTxSidecar) {
+	var (
+		txset      = make([][]*types.Transaction, len((*p).BlockBodies))
```
minh-bq, line 285: Can we use `p.BlockBodies` instead of `(*p).BlockBodies)` here?

eth/handler.go
```diff
@@ -544,7 +545,13 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 		// Send the block to a subset of our peers
 		transfer := peers[:int(math.Sqrt(float64(len(peers))))]
 		for _, peer := range transfer {
-			peer.AsyncSendNewBlock(block, td)
+			var sidecars []types.BlobTxSidecar
+			if peer.Version() >= eth.ETH100 {
+				for _, blobSidecar := range rawdb.ReadBlobSidecars(h.database, block.Hash(), block.NumberU64()) {
+					sidecars = append(sidecars, (*blobSidecar).BlobTxSidecar)
```
minh-bq, line 551: Can we use blobSidecar.BlobTxSidecar instead?

eth/protocols/eth/handlers.go
minh-bq, line 1: We need to extend the `answerGetBlockBodiesQuery` to answer the query with blob sidecars.

eth/protocols/eth/handler_test.go
minh-bq, line 1: We need to add some tests for ETH100 here (e.g. TestGetBlockBodies68)

eth/protocols/eth/handlers.go
Francesco4203: I've added `answerGetBlockBodiesQuery100`

eth/protocols/eth/handler_test.go
Francesco4203: I've added `TestGetBlockBodies100`

core/blockchain_reader.go
```diff
@@ -230,9 +230,9 @@ func (bc *BlockChain) GetBlobSidecarsByNumber(number uint64) *types.BlobSidecars
 	if sidecars, ok := bc.blobSidecarsCache.Get(hash); ok {
 		return sidecars.(*types.BlobSidecars)
 	}
-	
+
 	sidecars := rawdb.ReadBlobSidecars(bc.db, hash, number)
-	bc.blobSidecarsCache.Add(hash, sidecars)
+	bc.blobSidecarsCache.Add(hash, &sidecars)
```
minh-bq, line 235: Why do we need this change?

eth/protocols/eth/protocol.go
```diff
@@ -242,6 +277,20 @@ func (p *BlockBodiesPacket) Unpack() ([][]*types.Transaction, [][]*types.Header)
...
+		txset      = make([][]*types.Transaction, len((*p).BlockBodiesPacket))
+		uncleset   = make([][]*types.Header, len((*p).BlockBodiesPacket))
+		sidecarset = make([][]*types.BlobTxSidecar, len((*p).BlockBodiesPacket))
+	)
+	for i, body := range (*p).BlockBodiesPacket {
+		txset[i], uncleset[i], sidecarset[i] = body.Transactions, body.Uncles, (*p).Sidecars[i]
+	}
+	return txset, uncleset, sidecarset
+}
+
```
minh-bq, line 293: Can we use p.BlockBodiesPacket instead of (*p).BlockBodiesPacket and some other places in this function?

minh-bq (commented):
Some minor comments, overall LGTM

eth/protocols/eth/protocol.go
Francesco4203: fixed, I saw the previous code used p since the pointer can't be used directly, but I didn't know pointers can be dereferenced automatically if called to its attributes

core/blockchain_reader.go
Francesco4203: The `sidecars` is retrieved as follow, so it should be added as pointer, otherwise it can't be cast.
```
if sidecars, ok := bc.blobSidecarsCache.Get(hash); ok {
	return sidecars.(*types.BlobSidecars)
}
```

minh-bq (approved):

PR #517, opened 2024-08-07
blockchain: blob sidecar pruning

Francesco4203:


core/blockchain.go
```diff
@@ -1427,6 +1427,41 @@ func (bc *BlockChain) reorgNeeded(localBlock *types.Block, localTd *big.Int, ext
 	return reorg
 }
 
+// pruneBlockSidecars prunes the sidecars of blocks that are older than the keep period
+func (bc *BlockChain) pruneBlockSidecars(db ethdb.KeyValueWriter) {
```
minh-bq, line 1431: We should follow the specification in the design document:
"When inserting a new canonical block, prune the blobs of block at: lastest - 518,400 (around 18 days)"

core/rawdb/schema.go
```diff
@@ -92,7 +92,7 @@ var (
 	blockBodyPrefix     = []byte("b") // blockBodyPrefix + num (uint64 big endian) + hash -> block body
 	blockReceiptsPrefix = []byte("r") // blockReceiptsPrefix + num (uint64 big endian) + hash -> block receipts
 
-	blobSidecarsPrefix = []byte("s") // blobSidecarsPrefix + num (uint64 big endian) + hash -> sidecars
+	BlobSidecarsPrefix = []byte("s") // blobSidecarsPrefix + num (uint64 big endian) + hash -> sidecars
```
minh-bq, line 95: Why do we need to export this variable?

Francesco4203: At first I decided to use it to iterate back the blobs in db, but I forgot to redo it. Fixed.

core/blockchain.go
Francesco4203: Oh I missed that, updating. Btw do you think that we should also change the logic for skipping blob check, which is currently checking using the time of 18 days?

core/blockchain.go
```diff
@@ -1459,6 +1459,20 @@ func (bc *BlockChain) reorgNeeded(localBlock *types.Block, localTd *big.Int, ext
...
+func (bc *BlockChain) pruneBlockSidecars(db ethdb.KeyValueWriter, curBlock *types.Block) {
+	if curBlock.NumberU64() < uint64(params.BlobPrunePeriod) {
+		return
+	}
+	pruneBlockNumber := curBlock.NumberU64() - uint64(params.BlobPrunePeriod)
+	pruneBlock := bc.GetBlockByNumber(pruneBlockNumber)
+	if pruneBlock == nil {
+		return
+	}
+	pruneBlockHash := pruneBlock.Hash()
```
minh-bq, line 1472: I'm not sure if it is better but we can use bc.GetCanonicalHash to get the block hash

minh-bq (approved):
LGTM

core/blockchain.go
Francesco4203: It is cleaner I believe, fixed.

PR #519, opened 2024-08-08
core, internal, eth: add blob tx to eth_call and eth_estimategas rpc methods

andicrypt:


core/types/transaction_test.go
```diff
@@ -268,10 +271,20 @@ func TestTransactionCoding(t *testing.T) {
...
+		eip2930Signer = NewEIP2930Signer(common.Big1)
+		// londonSigner  = NewLondonSigner(common.Big1)
+		cancunSigner = NewCancunSigner(common.Big1)
 	)
 	for i := uint64(0); i < 500; i++ {
 		var txdata TxData
-		switch i % 6 {
+		switch i % 7 {
+		case 6:
+			signer = cancunSigner
```
minh-bq, line 283: Why don't we just change signer to always be cancunSigner?

eth/gasestimator/gasestimator.go
```diff
@@ -80,6 +80,15 @@ func Estimate(ctx context.Context, call core.Message, opts *Options, gasCap uint
 			}
 			available.Sub(available, call.Value())
 		}
+		if opts.Config.IsCancun(opts.Header.Number) {
+			blobGasPerBlob := new(big.Int).SetUint64(params.BlobTxBlobGasPerBlob)
+			blobBalanceUsage := new(big.Int).SetUint64(uint64(len(call.BlobHashes())))
+			blobBalanceUsage.Mul(blobBalanceUsage, blobGasPerBlob)
+			blobBalanceUsage.ModInverse(blobBalanceUsage, call.BlobGasFeeCap())
```
minh-bq, line 87: Why do we use ModInverse here?

internal/ethapi/api.go
```diff
@@ -468,6 +470,9 @@ func (s *PrivateAccountAPI) SendTransaction(ctx context.Context, args Transactio
 		s.nonceLock.LockAddr(args.from())
 		defer s.nonceLock.UnlockAddr(args.from())
 	}
+	if args.IsEIP4844() {
```
minh-bq, line 473: Why don't we support EIP 4844 here?

eth/gasestimator/gasestimator.go
andicrypt: Fix, it's mistaken coding.

internal/ethapi/transaction_args.go
```diff
@@ -239,7 +294,7 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, baseFee *big.Int) (t
 	if args.AccessList != nil {
 		accessList = *args.AccessList
 	}
-	msg := types.NewMessage(addr, args.To, 0, value, gas, gasPrice, gasFeeCap, gasTipCap, data, accessList, true)
+	msg := types.NewMessage(addr, args.To, 0, value, gas, gasPrice, gasFeeCap, gasTipCap, data, accessList, true, (*big.Int)(args.BlobFeeCap), args.BlobHashes)
```
minh-bq, line 297: What if user inputs args.BlobHashes != nil but args.BlobFeeCap == nil, does it lead to a panic?

andicrypt: Can you give more context for this

minh-bq (approved):

PR #522, opened 2024-08-13
ethclient, core: add blob retrieval interface to ethclient

andicrypt:


core/txpool/blobpool/config.go
```diff
@@ -25,13 +25,17 @@ type Config struct {
 	Datadir   string // Data directory containing the currently executable blobs
 	Datacap   uint64 // Soft-cap of database storage (hard cap is larger due to overhead)
 	PriceBump uint64 // Minimum price bump percentage to replace an already existing nonce
+
+	FakeDump bool // FakeDump flags used to bypass dumping record into limbo
```
minh-bq, line 29: Why do we need this config?

minh-bq (approved):

PR #523, opened 2024-08-15
consortium-v2: make the system transaction mismatch error more verbose

minh-bq:
Currently, the error when system transaction in the received block is different from the local calculation does not show the transaction in received block information. This makes it quite hard to debug. This commit adds these information to the error.

### DIFF
--- a/consensus/consortium/common/contract.go
+++ b/consensus/consortium/common/contract.go
@@ -494,13 +494,13 @@ func ApplyTransaction(msg types.Message, opts *ApplyTransactOpts) (err error) {
 		// The system transaction is the transaction that have the receiver is ConsortiumV2Contracts
 		actualTx := (*receivedTxs)[0]
 		if !bytes.Equal(signer.Hash(actualTx).Bytes(), expectedHash.Bytes()) {
-			return fmt.Errorf("expected tx hash %v, get %v, nonce %d, to %s, value %s, gas %d, gasPrice %s, data %s", expectedHash.String(), actualTx.Hash().String(),
-				expectedTx.Nonce(),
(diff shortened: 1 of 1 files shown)


consensus/consortium/common/contract.go
```diff
@@ -494,13 +494,13 @@ func ApplyTransaction(msg types.Message, opts *ApplyTransactOpts) (err error) {
...
 		if !bytes.Equal(signer.Hash(actualTx).Bytes(), expectedHash.Bytes()) {
-			return fmt.Errorf("expected tx hash %v, get %v, nonce %d, to %s, value %s, gas %d, gasPrice %s, data %s", expectedHash.String(), actualTx.Hash().String(),
-				expectedTx.Nonce(),
-				expectedTx.To().String(),
-				expectedTx.Value().String(),
-				expectedTx.Gas(),
-				expectedTx.GasPrice().String(),
-				hex.EncodeToString(expectedTx.Data()),
+			return fmt.Errorf(
+				"expected transaction: nonce %d, to %s, value %s, gas %d, gasPrice %s, data %s\n"+
```
andicrypt, line 498: should we also log tx hash here

minh-bq: The reason tx's hash is not included into the log is that transaction hash is a result of hash on the whole transaction which includes the signature. When mismatch happens, the local node finds out that the local computed tx is different from remote tx, but it can't compute the signature value of its local computed tx. So it cannot correctly compute the local computed tx hash.

minh-bq: added

andicrypt (approved):

PR #528, opened 2024-08-16
all: introduce blob transaction to Ronin

minh-bq:


core/blockchain.go
```diff
@@ -1414,9 +1459,27 @@ func (bc *BlockChain) reorgNeeded(localBlock *types.Block, localTd *big.Int, ext
 	return reorg
 }
 
+// pruneBlockSidecars prunes the sidecars of blocks that are older than the keep period
+func (bc *BlockChain) pruneBlockSidecars(db ethdb.KeyValueWriter, curBlock *types.Block) {
```
andicrypt, line 1463: I think we should think about this carefully. It might happen the case when a node sync data from other node, what if other node is malicious, they send ancient block together with its sidecars (self-saved), and broadcast to us, we do save these ancient sidecars, and actually persist here permanently here.

consensus/consortium/v2/consortium.go
```diff
@@ -203,6 +209,90 @@ func (c *Consortium) Author(header *types.Header) (common.Address, error) {
 	return header.Coinbase, nil
 }
 
+// VerifyBlobHeader verifies a block's header blob and corresponding sidecar, returning BlobSideCars
+func (c *Consortium) VerifyBlobHeader(block *types.Block, sidecars []*types.BlobTxSidecar) (error, *types.BlobSidecars) {
```
andicrypt, line 213: I find that we do not use output of *types.BlobSidecars whenever we use this function, so I think we can remove types.BlobSidecars{} reconstruction part here.
On the other hand, tx.Hash can be added when we do `writeBlockSidecars` before dump into db.

core/txpool/blobpool/limbo.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// update changes the block number under which a blob transaction is tracked. This
+// method should be used when a reorg changes a transaction's inclusion block.
+//
+// The method may log errors for various unexpected scenarios but will not return
+// any of it since there's no clear error case. Some errors may be due to coding
+// issues, others caused by signers mining MEV stuff or swapping transactions. In
+// all cases, the pool needs to continue operating.
+func (l *limbo) update(txhash common.Hash, block uint64) {
```
minh-bq, line 179: This function can be optimized, we want update the block information in limboBlob, so we may just update the data when the rlp encoded sized does not change without deleting the old one and inserting the new old. However, the complexity might out-weight the performance gain. Just note this here in case we find reorg case big performance overhead.

andicrypt (approved):

PR #536, opened 2024-08-20
miner: remove attached sidecars when commit new blob transaction

andicrypt:
Blob sidecars should be discarded when commit new blob transaction, creating sidecars-free new mined block.


minh-bq:
We should add a unit test for this case

andicrypt:
> We should add a unit test for this case

Added!

minh-bq (approved):

PR #537, opened 2024-08-20
eth/protocols: add nil sidecars checking to sanityCheck for NewBlockPacket100

Francesco4203:


eth/protocols/eth/protocol.go
```diff
@@ -212,6 +212,11 @@ func (request *NewBlockPacket100) sanityCheck() error {
 	if tdlen := request.TD.BitLen(); tdlen > 100 {
 		return fmt.Errorf("too large block TD: bitlen %d", tdlen)
 	}
+	for _, tx := range request.Block.Transactions() {
+		if tx.BlobTxSidecar() != nil {
+			return fmt.Errorf("block should not contain sidecars")
```
minh-bq, line 217: I think the error should be "transaction in mined block should not contain sidecars"

Francesco4203: fixed

andicrypt:
Can you add test case for this

minh-bq (approved):

andicrypt (approved):

PR #538, opened 2024-08-20
miner/worker: check if the transaction is evicted before committing

minh-bq:
There is a case where a blob transaction returned from txpool pending transaction query is evicted before commit. This leads to transaction's lazy resolve returns nil. This commit adds the nil check to ensure the transaction is not evicted before committing.

### DIFF
--- a/miner/worker_test.go
+++ b/miner/worker_test.go
@@ -646,6 +646,12 @@ func newCurrent(t *testing.T, signer types.Signer, chain *core.BlockChain) *envi
 	return env
 }
 
+type mockPool struct{}
+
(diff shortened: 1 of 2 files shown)


miner/worker_test.go
```diff
@@ -646,6 +646,12 @@ func newCurrent(t *testing.T, signer types.Signer, chain *core.BlockChain) *envi
 	return env
 }
 
+type mockPool struct{}
+
+func (m *mockPool) Get(hash common.Hash) *types.Transaction {
```
andicrypt, line 651: Nice!

andicrypt (approved):

Francesco4203 (approved):

PR #540, opened 2024-08-22
p2p, eth: create ENR filtering earlier before p2p server starts

minh-bq:
The race detector reports the following data race
```
WARNING: DATA RACE
Read at 0x00c0014e60b4 by goroutine 4291:
  github.com/ethereum/go-ethereum/core/forkid.newFilter.func1()
      github.com/ethereum/go-ethereum/core/forkid/forkid.go:163 +0xf1
  github.com/ethereum/go-ethereum/p2p.(*Server).setupDiscovery.func2()
      github.com/ethereum/go-ethereum/p2p/server.go:615 +0x1a2
  github.com/ethereum/go-ethereum/p2p/discover.(*Table).filterNode()
      github.com/ethereum/go-ethereum/p2p/discover/table.go:585 +0x456
  github.com/ethereum/go-ethereum/p2p/discover.(*Table).addSeenNodeSync()
      github.com/ethereum/go-ethereum/p2p/discover/table.go:548 +0x1c4
  github.com/ethereum/go-ethereum/p2p/discover.(*Table).addSeenNode.func1()
      github.com/ethereum/go-ethereum/p2p/discover/table.go:527 +0x47

Previous write at 0x00c0014e60b4 by main goroutine:
  github.com/ethereum/go-ethereum/core/forkid.newFilter()
      github.com/ethereum/go-ethereum/core/forkid/forkid.go:127 +0x1a4
  github.com/ethereum/go-ethereum/core/forkid.NewFilter()
      github.com/ethereum/go-ethereum/core/forkid/forkid.go:99 +0xf7
  github.com/ethereum/go-ethereum/eth/protocols/eth.StartENRFilter()
      github.com/ethereum/go-ethereum/eth/protocols/eth/discovery.go:62 +0xa9
  github.com/ethereum/go-ethereum/eth.(*Ethereum).Start()
      github.com/ethereum/go-ethereum/eth/backend.go:655 +0x3b
```
The ENR filter is created when Ethereum starts and is passed to p2p server, this is a write access on ENR filter. The p2p server uses this filter when discovering a peer, this is a read access. Currently, the p2p server starts before Ethereum starts, so peer discovery happens concurrently with Ethereum's start. Moreover, the read and write access do not synchronise with each other so the data race happens. This commit moves the ENR filter creation to Ethereum creation which happens before p2p server starts to fix the data race.

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/consensus/consortium"
+	"github.com/ethereum/go-ethereum/core/forkid"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/vote"
 
@@ -214,6 +215,8 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		go eth.blockchain.StartFinalityVoteMonitor()
 	}
 
+	StartENRFilter(eth.blockchain, eth.p2pServer)
+
 	// Rewind the chain in case of an incompatible config upgrade.
 	if compat, ok := genesisErr.(*params.ConfigCompatError); ok {
 		log.Warn("Rewinding chain to upgrade configuration", "err", compat)
@@ -351,6 +354,11 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		}
 	}
 	return eth, nil
+}
+
+func StartENRFilter(chain *core.BlockChain, p2p *p2p.Server) {
+	forkFilter := forkid.NewFilter(chain)
+	p2p.SetFilter(forkFilter)
 }
 
 // MakeEthApiBackend is used by MakeChain to create a minimal eth API backend for consortium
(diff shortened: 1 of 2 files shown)


eth/backend.go
```diff
@@ -214,6 +215,8 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		go eth.blockchain.StartFinalityVoteMonitor()
 	}
 
+	StartENRFilter(eth.blockchain, eth.p2pServer)
```
huyngopt1994, line 218: Could we keep StartENRFilter in discovery.go seem this function belongs to discovery logic?  only move out logic from Start to New 

```
eth. StartENRFilter(eth.blockchain, eth.p2pServer)
```

minh-bq: The problem is that in the New function, there is a variable named `eth`, so we cannot access the `eth` package in discovery.go

huyngopt1994: Renaming the variable can work, but seems we need to replace it many times.
```
err := n.openEndpoints()
```
The p2p server will start after this entry, so I think the PR will solve this issue. LGTM

huyngopt1994 (approved):

DNK90:
@minh-bq 
I think using your solution can fix the current issue, but it is not actually fixed the race condition.
instead of doing moving the function out, I think you should add mutex to for getting and setting `forkFilter`
So basically, in `SetFilter`, it should become:
```
func (srv *Server) SetFilter(f forkid.Filter) {
         srv.mtx.Lock()
         defer srv.mtx.Unlock()
	srv.forkFilter = f
}
```

and also add a GetFilter function
```
func (srv *Server) GetFilter(f forkid.Filter) forkid.Filter {
         srv.mtx.Lock()
         defer srv.mtx.Unlock()
	return srv.forkFilter
}
```
In any case that call srv.forkFilter directly, we will change to use `GetFilter`

PR #543, opened 2024-08-22
consensus, core: enable base fee after Venoki fork

minh-bq:
After London fork, we always set base fee to 0. This commit implements REP-0022: Enable base fee on Ronin
(https://github.com/ronin-chain/REPs/blob/main/REP-0022/REP-0022.md) to enable base fee calculation and transfer that fee to treasury after Venoki fork.

### DIFF
--- a/consensus/misc/eip1559/eip1559.go
+++ b/consensus/misc/eip1559/eip1559.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/consensus/misc"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/params"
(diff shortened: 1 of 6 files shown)


consensus/misc/eip1559/eip1559.go
```diff
@@ -47,48 +49,53 @@ func VerifyEip1559Header(config *params.ChainConfig, parent, header *types.Heade
 
 // CalcBaseFee calculates the basefee of the header.
 func CalcBaseFee(config *params.ChainConfig, parent *types.Header) *big.Int {
-	return big.NewInt(0)
+	blockNumber := new(big.Int).Add(parent.Number, common.Big1)
 
-	/*
-		We temporarily set the base fee to 0, the following logic may be enabled in
-		the future hardfork.
+	if !config.IsShanghai(blockNumber) {
```
phuctd95, line 54: Let's use a Ronin hardfork name instead of Shanghai. This feature isn't actually part of the Shanghai hardfork.

consensus/misc/eip1559/eip1559.go
```diff
@@ -47,48 +49,53 @@ func VerifyEip1559Header(config *params.ChainConfig, parent, header *types.Heade
...
-			baseFeeDelta := x.Div(y, baseFeeChangeDenominator)
-
-			return math.BigMax(
-				x.Sub(parent.BaseFee, baseFeeDelta),
-				common.Big0,
-			)
+		return x.Add(parent.BaseFee, baseFeeDelta)
+	} else {
+		// If the parent's base fee is at the minimum already, fast return the minimum base fee
+		minimumBaseFee := big.NewInt(params.InitialBaseFee)
```
phuctd95, line 85: I think we should have another parameter for the minimum base fee instead of using the InitialBaseFee.

minh-bq: fixed, PTAL

andicrypt (approved):

PR #553, opened 2024-08-30
all: introduce trie owner notion

Francesco4203:
This PR is to:
- Add trie owner notion to trie types and `prefetcher`. Currently, each trie is represented by its root hash. There will be a case when two accounts have exactly the same storage information, and thus the same root hash of storage trie. Adding the owner information helps distinguishing them correctly, and each account has its own storage trie now.
- Previously, in `(t *Trie) Commit()`, if the trie is clean (maybe committed earlier) and nothing-dirty-to-commit now, there might still be some resolved nodes (maybe retrieved earlier), but the tracer is `Reset()`, so it will lose the information regarding those nodes. This PR set the root to its hash to ensure that all nodes are released after `Commit()`.


huyngopt1994:
@Francesco4203 Could u put some description about the concept of trie owner here from go-ethereum team ?

Francesco4203:
> @Francesco4203 Could u put some description about the concept of trie owner here from go-ethereum team ?

@huyngopt1994 Added.

core/state/trie_prefetcher.go
```diff
@@ -51,13 +51,12 @@ type triePrefetcher struct {
...
 }
 
-// newTriePrefetcher
 func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePrefetcher {
 	prefix := triePrefetchMetricsPrefix + namespace
 	p := &triePrefetcher{
 		db:       db,
 		root:     root,
-		fetchers: make(map[common.Hash]*subfetcher), // Active prefetchers use the fetchers map
+		fetchers: make(map[string]*subfetcher), // Active prefetchers use the fetchers map
```
huyngopt1994, line 59:
```
func (p *triePrefetcher) trieID(owner common.Hash, root common.Hash) string {
	return string(append(owner.Bytes(), root.Bytes()...))
}
```

so the new key will be changed from the root to `string(append(owner.Bytes(), root.Bytes()...))`

huyngopt1994: https://github.com/ethereum/go-ethereum/pull/24750#discussion_r875594327

trie/utils.go
```diff
@@ -120,15 +146,20 @@ func (t *tracer) copy() *tracer {
...
 	for key := range t.delete {
 		delete[key] = struct{}{}
 	}
+	for key, val := range t.origin {
+		origin[key] = val
+	}
 	return &tracer{
 		insert: insert,
 		delete: delete,
+		origin: origin,
```
huyngopt1994, line 163: then we will have the other PR for tracking this map data ?

huyngopt1994 (approved):

huyngopt1994:
LGTM, let merge to next Implementing.

PR #554, opened 2024-08-30
core: capture opcode execution in subscriber only

minh-bq:
enableAdditionalChainEvent is true in subscriber only to publish additional information including opcode execution. Currently, in insertChain, we already enableAdditionalChainEvent the before capturing opcode execution. However, in system transaction, we always capture opcode execution by using bc.OpEvents without checking enableAdditionalChainEvent. This commit moves the enableAdditionalChainEvent check into bc.OpEvents so that opcode execution is captured only when enableAdditionalChainEvent is true.

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -444,17 +444,20 @@ func (bc *BlockChain) ReadDirtyAccounts(hash common.Hash) []*types.DirtyStateAcc
 }
 
 func (bc *BlockChain) OpEvents() []*vm.PublishEvent {
-	return []*vm.PublishEvent{
-		{
-			OpCodes: []vm.OpCode{
-				vm.CALL,
-				vm.DELEGATECALL,
-				vm.CREATE,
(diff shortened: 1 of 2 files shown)


core/blockchain_reader.go
```diff
@@ -444,17 +444,20 @@ func (bc *BlockChain) ReadDirtyAccounts(hash common.Hash) []*types.DirtyStateAcc
...
+					vm.CREATE,
+					vm.CREATE2,
+				},
+				Event: &InternalTransactionEvent{},
 			},
-			Event: &InternalTransactionEvent{},
-		},
+		}
 	}
+	return nil
```
huyngopt1994, line 460: As reading the source code, it will pass nil instead of empty array to bc.processor.Process, is it safe ?

huyngopt1994 (commented):
LGTM

huyngopt1994 (approved):

PR #558, opened 2024-09-05
txpool: ensure the max tip based on gas fee cap is high enough

minh-bq:
effectiveTip = min(tipcap, feecap - basefee), so the too low feecap can make the effectiveTip not enough to cover the pool's required tip. This commit adds a check to ensure feecap - minBaseFee is not lower the the pool's required tip.


andicrypt (approved):
LGTM

PR #559, opened 2024-09-05
internal, ethclient: change namespace of blob related RPC to ronin

andicrypt:
As blob related RPC is only specific to Ronin protocol, which are different from those of eth standard RPC, so this PR aims to differentiate and group blob related RPC to a new namespace `ronin`.


internal/ethapi/api.go
```diff
@@ -2220,6 +2202,32 @@ func (s *PublicNetAPI) Version() string {
...
 
+type PublicRoninAPI struct {
+	b Backend
+}
+
+func NewPublicRoninAPI(b Backend) *PublicRoninAPI {
+	return &PublicRoninAPI{b}
+}
+
+// GetBlobSidecarsByHash returns the requested list of blob sidecars based on block number.
```
minh-bq, line 2213: *GetBlobSidecarsByNumber

PR #560, opened 2024-09-05
core, trie: rework trie commiter

huyngopt1994:
Reference in https://github.com/ethereum/go-ethereum/pull/25320.
It's implementing PR prepare for next implementing Path based. 

this PR is submitted to adjust the commit procedure. Now all the dirty nodes of a trie will be encapsulated in a struct called nodeSet for return. Multiple nodeSets can be merged together as a MergedNodeSet. Eventually the MergedNodeSet will be submitted to in-memory database as a whole to represent the state transition from block to block.

### DIFF
--- a/trie/iterator.go
+++ b/trie/iterator.go
@@ -361,10 +361,16 @@ func (it *nodeIterator) resolveHash(hash hashNode, path []byte) (node, error) {
 			}
 		}
 	}
-	resolved, err := it.trie.resolveHash(hash, path)
-	return resolved, err
+	return it.trie.resolveHash(hash, path)
 }
-
+func (it *nodeIterator) resolveBlob(hash hashNode, path []byte) ([]byte, error) {
+	if it.resolver != nil {
+		if blob, err := it.resolver.Get(hash); err == nil && len(blob) > 0 {
+			return blob, nil
+		}
+	}
+	return it.trie.resolveBlob(hash, path)
+}
 func (st *nodeIteratorState) resolve(it *nodeIterator, path []byte) error {
 	if hash, ok := st.node.(hashNode); ok {
 		resolved, err := it.resolveHash(hash, path)
--- a/trie/iterator.go
+++ b/trie/iterator.go
@@ -363,14 +363,7 @@ func (it *nodeIterator) resolveHash(hash hashNode, path []byte) (node, error) {
 	}
 	return it.trie.resolveHash(hash, path)
 }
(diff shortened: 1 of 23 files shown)


trie/iterator.go
```diff
@@ -361,10 +361,16 @@ func (it *nodeIterator) resolveHash(hash hashNode, path []byte) (node, error) {
 			}
 		}
 	}
-	resolved, err := it.trie.resolveHash(hash, path)
-	return resolved, err
+	return it.trie.resolveHash(hash, path)
+}
+func (it *nodeIterator) resolveBlob(hash hashNode, path []byte) ([]byte, error) {
```
Francesco4203, line 366: When do we use this function?

trie/nodeset.go
```diff
@@ -0,0 +1,95 @@
...
+func NewWithNodeSet(set *NodeSet) *MergedNodeSet {
+	merged := NewMergedNodeSet()
+	merged.Merge(set)
+	return merged
+}
+
+// Merge merges the provided dirty nodes of a trie into the set. The assumption
+// is held that no duplicated set belonging to the same trie will be merged twice.
+func (set *MergedNodeSet) Merge(other *NodeSet) error {
+	_, present := set.sets[other.owner]
```
Francesco4203, line 89: Just concerning, account trie does not have owner (owner is empty), will this still work as expected?

trie/iterator.go
huyngopt1994: It's used by NodeBlob functionality, but seem we don't support it in the current logic, let me remove this.

trie/nodeset.go
huyngopt1994: As I understood, The account trie will have the empty hash for the owner, but the StorageTrie will use the addrHash for it
```
// OpenStorageTrie opens the storage trie of an account.
func (db *cachingDB) OpenStorageTrie(addrHash, root common.Hash) (Trie, error) {
	tr, err := trie.NewSecure(addrHash, root, db.db)
```
For account owner, the mergednodeset will be initiated and merged one time when comitting so it's safe for case duplicated.

```
	// Write the account trie changes, measuing the amount of wasted time
	var start time.Time
	if metrics.EnabledExpensive {
		start = time.Now()
	}
	root, nodeSet, err := s.trie.Commit(true)
	if err != nil {
		return common.Hash{}, err
	}
```

Francesco4203 (approved):

PR #561, opened 2024-09-05
p2p/enr: reduce allocation in Record.encode (#24034)

huyngopt1994:
+2 ( including sig and seq ), avoid reallocation when over capacity
```
// AppendElements appends the sequence number and entries to the given slice.
func (r *Record) AppendElements(list []interface{}) []interface{} {
	list = append(list, r.seq)
	for _, p := range r.pairs {
		list = append(list, p.k, p.v)
	}
	return list
}
```

### DIFF
--- a/p2p/enr/enr.go
+++ b/p2p/enr/enr.go
@@ -304,7 +304,7 @@ func (r *Record) AppendElements(list []interface{}) []interface{} {
 }
 
 func (r *Record) encode(sig []byte) (raw []byte, err error) {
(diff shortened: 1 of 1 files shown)


minh-bq:
Can you add the reference to original commit in go-ethereum to the commit message for easier lookup (something like this https://github.com/axieinfinity/ronin/pull/528/commits/2981c18e2f0cc31693ab8f36be4ad6e2add9db7d)?

PR #564, opened 2024-09-06
core, eth: rework snap sync

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/24898

### DIFF
--- a/les/downloader/statesync.go
+++ b/les/downloader/statesync.go
@@ -35,7 +35,7 @@ import (
 // a single data retrieval network packet.
 type stateReq struct {
 	nItems    uint16                    // Number of items requested for download (max is 384, so uint16 is sufficient)
-	trieTasks map[common.Hash]*trieTask // Trie node download tasks to track previous attempts
+	trieTasks map[string]*trieTask      // Trie node download tasks to track previous attempts
 	codeTasks map[common.Hash]*codeTask // Byte code download tasks to track previous attempts
 	timeout   time.Duration             // Maximum round trip time for this to complete
 	timer     *time.Timer               // Timer to fire when the RTT timeout expires
@@ -264,7 +264,7 @@ type stateSync struct {
 	sched  *trie.Sync         // State trie sync scheduler defining the tasks
 	keccak crypto.KeccakState // Keccak256 hasher to verify deliveries with
 
-	trieTasks map[common.Hash]*trieTask // Set of trie node tasks currently queued for retrieval
+	trieTasks map[string]*trieTask      // Set of trie node tasks currently queued for retrieval
 	codeTasks map[common.Hash]*codeTask // Set of byte code tasks currently queued for retrieval
 
 	numUncommitted   int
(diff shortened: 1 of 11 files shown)


huyngopt1994:
```
	var onSlot func(paths [][]byte, hexpath []byte, leaf []byte, parent common.Hash) error
	var onSlot func(keys [][]byte, path []byte, leaf []byte, parent common.Hash, parentPath []byte) error
```
The reason behind why we changed from paths to keys and hexpath to path 
```
keys is a tuple of 32 bytes key. path is the hex format nibbles. Just want to make it less-ambiguous. 
```

The keys []byte will convert to hex format nibbles via this function 
```
func keybytesToHex(str []byte) []byte {
	l := len(str)*2 + 1
	var nibbles = make([]byte, l)
	for i, b := range str {
		nibbles[i*2] = b / 16
		nibbles[i*2+1] = b % 16
	}
	nibbles[l-1] = 16
	return nibbles
}
```

trie/sync.go
```diff
@@ -85,30 +73,57 @@ func NewSyncPath(path []byte) SyncPath {
...
+// codeRequest represents a scheduled or already in-flight bytecode retrieval request.
+type codeRequest struct {
+	hash    common.Hash    // Hash of the contract bytecode to retrieve
+	path    []byte         // Merkle path leading to this node for prioritization
+	data    []byte         // Data content of the node, cached until all subtrees complete
+	parents []*nodeRequest // Parent state nodes referencing this entry (notify all upon completion)
+}
+
+// NodeSyncResult is a response with requested trie node along with its node path.
+type NodeSyncResult struct {
```
huyngopt1994, line 96: split request logic into node and codem, same with SyncResult

les/downloader/statesync.go
```diff
@@ -586,13 +585,35 @@ func (s *stateSync) process(req *stateReq) (int, error) {
...
 	s.keccak.Reset()
 	s.keccak.Write(blob)
-	s.keccak.Read(res.Hash[:])
-	err := s.sched.Process(res)
-	return res.Hash, err
+	s.keccak.Read(hash[:])
+
+	if _, present := codeTasks[hash]; present {
+		err := s.sched.ProcessCode(trie.CodeSyncResult{
+			Hash: hash,
```
huyngopt1994, line 600: In the previous logic. 
```
err := s.sched.Process(res)
```
They will forward res to Process function , in the new logic, we need to split into handle either ProcessCode or ProcessNode. Could u explain a little the condition check here for choosing which kind of processing in sched.

Francesco4203: Since the requests list are also now divided into `nodeReqs` and `codeReqs`, we use `ProcessNode` for each request taking from `nodeReqs` and `ProcessCode` for each request taking from `codeReqs`

huyngopt1994:
LGTM, let merge and retest again

huyngopt1994 (approved):

PR #567, opened 2024-09-09
all: rework genesis api

Francesco4203:
This PR refers to the 2 following PRs:
- https://github.com/ethereum/go-ethereum/pull/24460: allow recommitting genesis state in case of rewinding to it. The current code assume genesis state is always in the database, but PBSS might overwrite the old states.
- https://github.com/ethereum/go-ethereum/pull/25473: rework a few things:
- - `ToBlock` is now only used to generate genesis block, since we have `Flush` to write to db
- - Remove `GenesisBlockForTesting`

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -80,6 +80,81 @@ func (ga *GenesisAlloc) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// flush adds allocated genesis accounts into a fresh new statedb and
+// commit the state changes into the given database handler.
+func (ga *GenesisAlloc) flush(db ethdb.Database) (common.Hash, error) {
+	statedb, err := state.New(common.Hash{}, state.NewDatabase(db), nil)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	for addr, account := range *ga {
+		statedb.AddBalance(addr, account.Balance)
+		statedb.SetCode(addr, account.Code)
+		statedb.SetNonce(addr, account.Nonce)
+		for key, value := range account.Storage {
+			statedb.SetState(addr, key, value)
+		}
+	}
+	root, err := statedb.Commit(false)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	err = statedb.Database().TrieDB().Commit(root, true, nil)
+	if err != nil {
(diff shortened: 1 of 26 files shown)


huyngopt1994:
Founding some places impacted to rewind logic here, maybe need to retest again after merging. 
* SetHead
* Head State Missing

core/genesis.go
```diff
@@ -80,6 +80,81 @@ func (ga *GenesisAlloc) UnmarshalJSON(data []byte) error {
...
+			genesis = DefaultGenesisBlock()
+		case params.RopstenGenesisHash:
+			genesis = DefaultRopstenGenesisBlock()
+		case params.RinkebyGenesisHash:
+			genesis = DefaultRinkebyGenesisBlock()
+		case params.GoerliGenesisHash:
+			genesis = DefaultGoerliGenesisBlock()
+		case params.SepoliaGenesisHash:
+			genesis = DefaultSepoliaGenesisBlock()
+		}
```
huyngopt1994, line 163: Could u recheck all params when recommit genesis to disk?

core/genesis.go
```diff
@@ -262,25 +352,12 @@ func (g *Genesis) configOrDefault(ghash common.Hash) *params.ChainConfig {
...
 	}
-	for addr, account := range g.Alloc {
-		statedb.AddBalance(addr, account.Balance)
-		statedb.SetCode(addr, account.Code)
-		statedb.SetNonce(addr, account.Nonce)
-		for key, value := range account.Storage {
-			statedb.SetState(addr, key, value)
-		}
-	}
-	root := statedb.IntermediateRoot(false)
```
huyngopt1994, line 283: Is it safe to remove the recalculated root in this line?

huyngopt1994:
```
	root, err := g.Alloc.flush(db)
	if err != nil {
		panic(err)
	}
```
The root is calculated by flush logic which is committed, so we don't need to callI IntermediateRoot here 🤔

huyngopt1994:
I rechecked the logic of refactoring, seem it's safe for merging. LGTM.

huyngopt1994 (approved):

PR #570, opened 2024-09-11
all: move genesis init to blockchain

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/25523/commits

Move genesis initialization logic to `NewBlockChain`:
- Current flow: `backend.New()` calls `SetupGenesisBlockWithOverride`, which does open `Trie.db` for initializing. Then, it calls `NewBlockChain`, which also opens `Trie.db` for progressing the chain. 
- - This is acceptable for now, since `Trie.db` can be opened multiple times. However, later in PBSS, `Trie.db` contains diff freezer, which is a singleton that can only be open once (and thus must be closed before reopening). Moving all the use of `Trie.db` in the same component makes it easier for closing logic.
- New flow: `SetupGenesisBlockWithOverride` is called inside `NewBlockChain`.
- - Later PRs will open `Trie.db` once within the `BlockChain` component, and then close it in `BlockChain.Close()`.

Note that the later `Trie.db` that contains diff freezer will be reworked (and thus renamed), we just use its current name here for convenience.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2223,7 +2223,8 @@ func MakeGenesis(ctx *cli.Context) *core.Genesis {
 func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chainDb ethdb.Database) {
 	var err error
 	chainDb = MakeChainDatabase(ctx, stack, false) // TODO(rjl493456442) support read-only database
-	config, _, err := core.SetupGenesisBlock(chainDb, MakeGenesis(ctx), false)
+	gpec := MakeGenesis(ctx)
+	config, err := core.LoadChainConfig(chainDb, gpec)
 	if err != nil {
 		Fatalf("%v", err)
 	}
@@ -2282,7 +2283,7 @@ func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chai
 
 	// TODO(rjl493456442) disable snapshot generation/wiping if the chain is read only.
 	// Disable transaction indexing/unindexing by default.
-	chain, err = core.NewBlockChain(chainDb, cache, config, engine, vmcfg, nil, nil)
+	chain, err = core.NewBlockChain(chainDb, cache, gpec, nil, engine, vmcfg, nil, nil)
 	if err != nil {
 		Fatalf("Can't create BlockChain: %v", err)
 	}
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -136,11 +136,6 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if err != nil {
 		return nil, err
 	}
-	chainConfig, genesisHash, genesisErr := core.SetupGenesisBlockWithOverride(chainDb, config.Genesis, config.OverrideArrowGlacier, false)
-	if _, ok := genesisErr.(*params.ConfigCompatError); genesisErr != nil && !ok {
-		return nil, genesisErr
-	}
-	log.Info("Initialised chain configuration", "config", chainConfig)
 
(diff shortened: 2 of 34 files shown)


params/config.go
```diff
@@ -476,6 +476,36 @@ var (
 		Consortium:                    nil,
 		ConsortiumV2Contracts:         nil,
 	}
+	NonActivatedConfig = &ChainConfig{
```
Francesco4203, line 479: Moving genesis init to `NewBlockChain` results in more checking when calling it. `TestChainConfig` currently turns on a lot of fork blocks at block 0, which makes some tests to fail. `NonActivatedConfig` is added with all features turned off so that we can adjust manually for each test.

eth/backend.go
```diff
@@ -249,7 +242,7 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 		if _, err := vote.NewVoteManager(
 			eth,
 			chainDb,
-			chainConfig,
+			eth.blockchain.Config(),
```
huyngopt1994, line 245: @Francesco4203 it's better u asign an variable 
chainConfig := eth.blockchain.Config(), then pass it for avoiding changing so much ?

huyngopt1994: same with genesisHash

cmd/utils/flags.go
```diff
@@ -2223,7 +2223,8 @@ func MakeGenesis(ctx *cli.Context) *core.Genesis {
 func MakeChain(ctx *cli.Context, stack *node.Node) (chain *core.BlockChain, chainDb ethdb.Database) {
 	var err error
 	chainDb = MakeChainDatabase(ctx, stack, false) // TODO(rjl493456442) support read-only database
-	config, _, err := core.SetupGenesisBlock(chainDb, MakeGenesis(ctx), false)
+	gpec := MakeGenesis(ctx)
+	config, err := core.LoadChainConfig(chainDb, gpec)
```
huyngopt1994, line 2227: hmm why we need the change here ?

params/config.go
huyngopt1994: @Francesco4203 so as I understood, moving this logic will make some test trigger the logic genesis init step here, then  when using param.TestChainConfig it will be compatiable with current test, then we need to reinit new config for testing right ? Why does we have the the custom failed here comparing to  PR in ethereum, and turning off cnfig will skip some testing checking or not ?

cmd/utils/flags.go
Francesco4203: Same logic in `backend`, there is a line below that calls `core.NewBlockChain`:
```
chain, err = core.NewBlockChain(chainDb, cache, gpec, nil, engine, vmcfg, nil, nil)
```

params/config.go
Francesco4203: @huyngopt1994 
- Yes.

>  Why does we have the the custom failed here comparing to PR in ethereum
- What custom failed are you referring to?
>  and turning off cnfig will skip some testing checking or not ?
- For now, this config is only used in `core/dao_test.go/TestDAOForkRangeExtradata`, in which we turning off all configs and only turn on the DAO fork block. The reason for this is that in that test case, the DAO fork block is set at  32, not 0 as in the `TestChainConfig`, so we have to set it manually. The checking in genesis init will still happen. Not sure if this `NonActivatedConfig` is going to be used for more test in the future.

huyngopt1994: > What custom failed are you referring to?

Ah I means that Why does Ethereum encounter this issue like us. Any reason behind

huyngopt1994: Ok I found some logics changed from the original PR.

eth/backend.go
Francesco4203: Agree, fixed

huyngopt1994 (approved):

PR #575, opened 2024-09-17
core/rawdb: fix double-lock causing hang (#24189)

minh-bq:
commit https://github.com/ethereum/go-ethereum/commit/66a908c5e87a4f38c0507ae2c7e53b242deb7128.

Recursive read-locking is prohibited in RWMutex which can lead to deadlock. ReadCanonicalHash internally calls ReadAncients which cause recursive read-locking. This commit reads the block directly from leveldb/pebbledb in case block is not in ancient without using helper function ReadCanonicalHash.


tien-nguyen-sm (approved):
LGTM

andicrypt (approved):

PR #576, opened 2024-09-17
cmd, core, eth, trie: track deleted nodes

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/25757

Important changes:

- A storage trie is now identified by StateRoot (root of state/account trie), Owner, Root. The state/account trie is identified with StateRoot = Root. Those params are now needed for init a trie.

- Separate list of dirty nodes to updated nodes and deleted node. Also separate the updating/deleting operations.

- Track previous value of each node (before committing).

- Apply delete tracking logics to committer.

- Add `trie_reader.go`, interface for reaching database layer(s) from trie. Support backward compatibility: can use `hash_reader` for old hash based scheme, and can use `diffLayer` for path based scheme in the future PRs.

- Add metrics for tracking trie commits.

### DIFF
--- a/trie/nodeset.go
+++ b/trie/nodeset.go
@@ -18,52 +18,171 @@ package trie
 
 import (
 	"fmt"
+	"reflect"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // memoryNode is all the information we know about a single cached trie node
 // in the memory.
 type memoryNode struct {
-	hash common.Hash // Node hash, computed by hashing rlp value
-	size uint16      // Byte size of the useful cached data
-	node node        // Cached collapsed trie node, or raw rlp data
+	hash common.Hash // Node hash, computed by hashing rlp value, empty for deleted nodes
+	size uint16      // Byte size of the useful cached data, 0 for deleted nodes
+	node node        // Cached collapsed trie node, or raw rlp data, nil for deleted nodes
+}
+
+// memoryNodeSize is the raw size of a memoryNode data structure without any
+// node data included. It's an approximate size, but should be a lot better
+// than not counting them.
+// nolint:unused
+var memoryNodeSize = int(reflect.TypeOf(memoryNode{}).Size())
+
+// memorySize returns the total memory size used by this node.
+// nolint:unused
+func (n *memoryNode) memorySize(key int) int {
+	return int(n.size) + memoryNodeSize + key
+}
+
+// rlp returns the raw rlp encoded blob of the cached trie node, either directly
+// from the cache, or by regenerating it from the collapsed node.
+// nolint:unused
+func (n *memoryNode) rlp() []byte {
+	if node, ok := n.node.(rawNode); ok {
+		return node
+	}
+	enc, err := rlp.EncodeToBytes(n.node)
+	if err != nil {
+		log.Error("Failed to encode trie node", "err", err)
+	}
+	return enc
+}
+
+// obj returns the decoded and expanded trie node, either directly from the cache,
+// or by regenerating it from the rlp encoded blob.
+// nolint:unused
+func (n *memoryNode) obj() node {
+	if node, ok := n.node.(rawNode); ok {
+		return mustDecodeNode(n.hash[:], node)
+	}
+	return expandNode(n.hash[:], n.node)
+}
+
+// nodeWithPrev wraps the memoryNode with the previous node value.
+type nodeWithPrev struct {
+	*memoryNode
+	prev []byte // RLP-encoded previous value, nil means it's non-existent
+}
+
+// unwrap returns the internal memoryNode object.
+// nolint:unused
+func (n *nodeWithPrev) unwrap() *memoryNode {
+	return n.memoryNode
+}
+
+// memorySize returns the total memory size used by this node. It overloads
+// the function in memoryNode by counting the size of previous value as well.
+// nolint: unused
+func (n *nodeWithPrev) memorySize(key int) int {
+	return n.memoryNode.memorySize(key) + len(n.prev)
+}
+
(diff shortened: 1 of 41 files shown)


trie/trie.go
```diff
@@ -563,19 +566,22 @@ func concat(s1 []byte, s2 ...byte) []byte {
...
-func (t *Trie) resolveHash(n hashNode, prefix []byte) (node, error) {
-	hash := common.BytesToHash(n)
-	if node := t.db.node(hash); node != nil {
-		return node, nil
+// resolveAndTrack loads node from the underlying store with the given node hash
+// and path prefix and also tracks the loaded node blob in tracer treated as the
+// node's original value. The rlp-encoded blob is preferred to be loaded from
+// database because it's easy to decode node while complex to encode node to blob.
+func (t *Trie) resolveAndTrack(n hashNode, prefix []byte) (node, error) {
+	blob, err := t.reader.nodeBlob(prefix, common.BytesToHash(n))
```
Francesco4203, line 579: We need to track the encoded node here, so get the blob first for tracer and then return the decoded one.

trie/utils.go
```diff
@@ -111,19 +111,36 @@ func (t *tracer) deleteList() [][]byte {
...
+		paths = append(paths, []byte(path))
+		blobs = append(blobs, blob)
+	}
+	return paths, blobs
+}
+
 // getPrev returns the cached original value of the specified node.
-func (t *tracer) getPrev(key []byte) []byte {
+func (t *tracer) getPrev(path []byte) []byte {
 	// Don't panic on uninitialized tracer, it's possible in testing.
```
huyngopt1994, line 139: // Tracer isn't used right now, remove this check later. -> Update it for correcting the status.

trie/utils_test.go
```diff
@@ -122,3 +123,123 @@ func TestTrieTracerNoop(t *testing.T) {
...
+	// Ensure all nodes are tracked by tracer with correct prev-values
+	iter := trie.NodeIterator(nil)
+	seen := make(map[string][]byte)
+	for iter.Next(true) {
+		// Embedded nodes are ignored since they are not present in
+		// database.
+		if iter.Hash() == (common.Hash{}) {
+			continue
+		}
+		blob, err := trie.reader.nodeBlob(iter.Path(), iter.Hash())
```
huyngopt1994, line 176: `seen[string(iter.Path())] = common.CopyBytes(iter.NodeBlob()) I saw they use iter.NodeBlob()` here ? Do we implement it ?

trie/trie_test.go
```diff
@@ -443,18 +446,50 @@ func runRandTest(rt randTest) bool {
...
+			}
+			_, err = VerifyProof(hash, step.key, proofDb)
+			if err != nil {
+				rt[i].err = fmt.Errorf("failed for verifying key %#x, %v", step.key, err)
+			}
 		case opHash:
 			tr.Hash()
 		case opCommit:
-			hash, nodes, err := tr.Commit(false)
+			root, nodes, err := tr.Commit(true)
```
huyngopt1994, line 466: Not sure why they change to collect Dirty Leaf here too  🤔

trie/utils_test.go
Francesco4203: Source: https://github.com/ethereum/go-ethereum/commit/55430b6ea2bcc63bc11e7089690d90d8ac70dc5c
We haven't implemented this yet. They said that it is needed for path-based, but I'm not sure. Until now we can still get the node blob like this.

trie/trie_test.go
```diff
@@ -443,18 +446,50 @@ func runRandTest(rt randTest) bool {
...
+						return false
+					}
+				}
+			}
 			if nodes != nil {
 				triedb.Update(NewWithNodeSet(nodes))
 			}
-			newtr, err := New(common.Hash{}, hash, triedb)
+			newtr, err := New(TrieID(root), triedb)
 			if err != nil {
```
huyngopt1994, line 493:
```
			// Enable node tracing. Resolve the root node again explicitly
			// since it's not captured at the beginning.
			tr.tracer = newTracer()
			tr.resolveAndTrack(root.Bytes(), nil)

			origTrie = tr.Copy()
```
Could u verify the logic enabling tracer and resolve  the root node here ?

trie/nodeset.go
```diff
@@ -18,52 +18,167 @@ package trie
...
+}
+
+// rlp returns the raw rlp encoded blob of the cached trie node, either directly
+// from the cache, or by regenerating it from the collapsed node.
+// nolint:unused
+func (n *memoryNode) rlp() []byte {
+	if node, ok := n.node.(rawNode); ok {
+		return node
+	}
+	enc, _ := rlp.EncodeToBytes(n.node)
```
huyngopt1994, line 55: hm Could u track error here ?

trie/utils_test.go
Francesco4203: Okay it is needed. I'll add it later.

trie/nodeset.go
Francesco4203: I remember they say that EncodeToBytes is expected to never return error, since we can always encode anything, and many later updates also do not track error here. Not sure if it is needed.

huyngopt1994: hm when I search rlp.EncodeToBytes, I see they checks for everytimes, Fatal or even panic incase getting error, for our case I think we just need to track Error here. At least we had some thing for tracing  🤔

Francesco4203: Hmm okay I agree, error checking added

trie/trie_test.go
Francesco4203: I think since they've just enabled the tracer here, so they just try to commit everything for full coverage.

trie/trie_test.go
Francesco4203: I missed that part in the test, added.

huyngopt1994 (approved):
then LGTM.

PR #578, opened 2024-09-18
 core,eth,tests,trie: abstract node scheme, and contruct database

huyngopt1994:
interface instead of keyvalue for supporting storing diff reverse data in ancient

Reference https://github.com/ethereum/go-ethereum/pull/25532

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -175,6 +175,7 @@ type BlockChain struct {
 	snaps  *snapshot.Tree // Snapshot tree for fast trie leaf access
 	triegc *prque.Prque   // Priority queue mapping block numbers to tries to gc
 	gcproc time.Duration  // Accumulates canonical block processing for trie dumping
+	triedb *trie.Database // The database handler for maintaining trie nodes.
 
 	// txLookupLimit is the maximum number of blocks from head whose tx indices
 	// are reserved:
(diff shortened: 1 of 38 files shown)


trie/stacktrie.go
```diff
@@ -242,26 +245,26 @@ func (st *StackTrie) getDiffIndex(key []byte) int {
 }
 
 // Helper function to that inserts a (key, value) pair into
-// the trie.
-func (st *StackTrie) insert(key, value []byte) {
+// the trie. Adding the prefix when inserting too.
+func (st *StackTrie) insert(key, value []byte, prefix []byte) {
```
Francesco4203, line 249: So as I understand here, the interface for inserting (key, value) pairs requires also the path. So the `prefix` here is the path from root to the current node, and it is used for `hash()` where we call function to insert the (key, value) pair. Is it correct?

core/blockchain.go
```diff
@@ -175,6 +175,7 @@ type BlockChain struct {
 	snaps  *snapshot.Tree // Snapshot tree for fast trie leaf access
 	triegc *prque.Prque   // Priority queue mapping block numbers to tries to gc
 	gcproc time.Duration  // Accumulates canonical block processing for trie dumping
+	triedb *trie.Database // The database handler for maintaining trie nodes.
```
Francesco4203, line 178: Can you explain why we are adding this? Previously, we use `bc.stateCache.TrieDB()`, so are they different? If so, then, is the logic changed here?

trie/sync.go
```diff
@@ -539,3 +539,14 @@ func (s *Sync) commitCodeRequest(req *codeRequest) error {
 	}
 	return nil
 }
+
+// ResolvePath resolves the provided composite node path by separating the
+// path in account trie if it's existent.
+func ResolvePath(path []byte) (common.Hash, []byte) {
```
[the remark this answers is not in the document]
huyngopt1994: I saw they use it for resolve path in https://github.com/ethereum/go-ethereum/blob/master/trie/sync.go#L575 future implementing.

core/blockchain.go
huyngopt1994: As I understood, it just move out the triedb object from stateCache obj to blockchain obj, behind they still point to same db same struct and config.

huyngopt1994: one reason, maybe they wanna decouple triedb object from cache state for easily feature enhancement 🤔

trie/stacktrie.go
huyngopt1994: and it is used for hash() where we call function to insert the (key, value) pair. Is it correct?-> It will use when u wanna flush to disk when u wanna insert and hash
-> It is the path from root to the current node, when start insert, path is nil 
```	
if st.writeFn != nil {
		st.writeFn(st.owner, path, common.BytesToHash(st.val), h.tmp)
	}
```

huyngopt1994: One example, how it store in. state/snapshot 
```

	var nodeWriter trie.NodeWriteFunc
	// Implement nodeWriter in case db is existed otherwise let it be nil.
	if db != nil {
		nodeWriter = func(owner common.Hash, path []byte, hash common.Hash, blob []byte) {
			scheme.WriteTrieNode(db, owner, path, hash, blob)
		}
	}
	t := trie.NewStackTrieWithOwner(nodeWriter, owner)
```

Francesco4203: Ok I got it.

Francesco4203 (approved):

huyngopt1994:
I will make a commit for enabling logic test `tests/fuzzers/stacktrie/trie_fuzzer.go` in tracked delete nodes PR.

PR #582, opened 2024-09-25
all: prep for path-based trie storage

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/26603

Changes: 
- Move trie storage logic to `accessors_trie.go`; separate the operations into old hash based (for backward compatibility) and new path based.
- Delete `NodeScheme`. `scheme` is no longer an interface, but just a string, which will be checked by the new logic in `accessors_trie.go` to choose the corresponding operations.


core/state/pruner/pruner.go
```diff
@@ -265,7 +265,7 @@ func (p *Pruner) Prune(root common.Hash) error {
 	// Ensure the root is really present. The weak assumption
 	// is the presence of root can indicate the presence of the
 	// entire trie.
-	if blob := rawdb.ReadTrieNode(p.db, root); len(blob) == 0 {
+	if blob := rawdb.ReadLegacyTrieNode(p.db, root); len(blob) == 0 {
```
huyngopt1994, line 268: -> `if !rawdb.HasLegacyTrieNode(p.db, root)` Could we change to this logic ? hm no need for reading data here right ?

les/downloader/downloader.go
```diff
@@ -229,7 +229,7 @@ func New(checkpoint uint64, stateDb ethdb.Database, stateBloom *trie.SyncBloom,
 		headerProcCh:   make(chan []*types.Header, 1),
 		quitCh:         make(chan struct{}),
 		stateCh:        make(chan dataPack),
-		SnapSyncer:     snap.NewSyncer(stateDb, nil),
+		SnapSyncer:     snap.NewSyncer(stateDb, ""),
```
huyngopt1994, line 232: hm ```SnapSyncer:     snap.NewSyncer(stateDb, rawdb.HashScheme),``` As the previous PR, it points to HashScheme ?

huyngopt1994: Do we miss update `les/downloader/statesync.go`

trie/sync.go
```diff
@@ -194,7 +194,8 @@ func (s *Sync) AddSubTrie(root common.Hash, path []byte, parent common.Hash, par
 		// Bloom filter says this might be a duplicate, double check.
 		// If database says yes, then at least the trie node is present
 		// and we hold the assumption that it's NOT legacy contract code.
-		blob := rawdb.ReadTrieNode(s.database, root)
+		owner, inner := ResolvePath(path)
+		blob := rawdb.ReadTrieNode(s.database, owner, inner, root, s.scheme)
```
huyngopt1994, line 198:
```
if rawdb.HasTrieNode(s.database, owner, inner, root, s.scheme) {
     return
}
```

les/downloader/downloader.go
Francesco4203: As I checked, it was previously nil, and we will not use les anw, so we might skip this?

core/state/pruner/pruner.go
Francesco4203: You're right. Updated.

les/downloader/downloader.go
huyngopt1994: Oke then let keep it like it, will comeback in case any issues happens.

huyngopt1994 (approved):
LGTM

PR #585, opened 2024-09-26
core, trie: rework trie database

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/26813

This PR generalize the trie database for later use of Verkle Tree

Some changes:

- database.go: `cacheNode` no longer stores node as a whole node, but the encoded bytes instead
- database.go: move resolver logic to committer, since it is specific for MPT
- trie_reader.go: only provide retrieving encoded node now


huyngopt1994 (commented):
LGTM

PR #589, opened 2024-09-30
trie, core: track state changes in statedb

Francesco4203:
Reference: https://github.com/ethereum/go-ethereum/pull/27349

This PR is for tracking state changes in stateDB. The tracked changes will later be used in PBSS to generate the history data.

Changes:
- trie/triestate/state.go: create struct `Set` representing the state changes that will be tracked. There will be pairs of values before and after being changed.
- trie/database_wrap: change name to database; `database.Update()` now take in the tracked state changes, which will later be used in PBSS.
- core/state_object: `stateObject` represents an account being updated. Now, `origin`, which represents the account before changes, is added for keeping track of changes. This `origin` will be updated to the current account only after committing.
- core/statedb: 
- - accounts/storages origins and changes are always tracked now (previously they are tracked only if snap is enabled).
- - `stateObjectsDestruct` that store destructed objects will be stored along with their previous values.
- - Add `handleDestruction` to track the destructed objects
- - All the changes (updates and deletes) will be stored to `Set` and passed to `s.db.TrieDB().Update()` when committing.

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -90,8 +90,18 @@ type (
 		account *common.Address
 	}
 	resetObjectChange struct {
+		account      *common.Address
 		prev         *stateObject
 		prevdestruct bool
+
+		// tracking previous states of accounts and storages in snapshot, before each transaction
+		prevAccount []byte
+		prevStorage map[common.Hash][]byte
+
+		// tracking previous states of accounts and storages in trie, before each commit
+		prevAccountOriginExist bool
+		prevAccountOrigin      []byte
+		prevStorageOrigin      map[common.Hash][]byte
 	}
 	selfDestructChange struct {
 		account     *common.Address
@@ -157,12 +167,21 @@ func (ch createObjectChange) dirtied() *common.Address {
 func (ch resetObjectChange) revert(s *StateDB) {
 	s.setStateObject(ch.prev)
 	if !ch.prevdestruct && s.snap != nil {
-		delete(s.snapDestructs, ch.prev.addrHash)
+		delete(s.stateObjectsDestruct, ch.prev.address)
+	}
+	if ch.prevAccountOriginExist {
+		s.accountsOrigin[ch.prev.addrHash] = ch.prevAccountOrigin
+	}
+	if ch.prevAccount != nil {
+		s.accounts[ch.prev.addrHash] = ch.prevAccount
+	}
+	if ch.prevStorage != nil {
+		s.storages[ch.prev.addrHash] = ch.prevStorage
 	}
 }
 
 func (ch resetObjectChange) dirtied() *common.Address {
-	return nil
+	return ch.account
 }
 
 func (ch selfDestructChange) revert(s *StateDB) {
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -35,6 +35,7 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
 	"github.com/ethereum/go-ethereum/trie/trienode"
+	"github.com/ethereum/go-ethereum/trie/triestate"
 )
 
 type revision struct {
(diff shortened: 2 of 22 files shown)


core/state/journal.go
```diff
@@ -90,8 +90,14 @@ type (
...
 	}
 	resetObjectChange struct {
-		prev         *stateObject
-		prevdestruct bool
+		account                *common.Address
+		prev                   *stateObject
+		prevdestruct           bool
+		prevAccount            []byte
+		prevStorage            map[common.Hash][]byte
+		prevAccountOriginExist bool
```
huyngopt1994, line 98: Could u split space here between Account & AccountOrigin and put some small documentation here for explaining the diferrent tracking here.

core/state/statedb.go
```diff
@@ -73,16 +74,21 @@ type StateDB struct {
...
+	storagesOrigin map[common.Hash]map[common.Hash][]byte // The original value of mutated slots in prefix-zero trimmed rlp format
 
 	// This map holds 'live' objects, which will get modified while processing a state transition.
-	stateObjects        map[common.Address]*stateObject
-	stateObjectsPending map[common.Address]struct{} // State objects finalized but not yet written to the trie
-	stateObjectsDirty   map[common.Address]struct{} // State objects modified in the current execution
+	stateObjects         map[common.Address]*stateObject
+	stateObjectsPending  map[common.Address]struct{}            // State objects finalized but not yet written to the trie
+	stateObjectsDirty    map[common.Address]struct{}            // State objects modified in the current execution
+	stateObjectsDestruct map[common.Address]*types.StateAccount // State objects destructed in the block along with its previous value
```
huyngopt1994, line 91: hm which PR migrated snapDestructs -> stateObjectsDestruct ?

huyngopt1994:
We have an interested  thing here, need to track latter, In case we don't track dirties account in journal resetobject entry when we encounter create the existing account. The test `state_fuzz_test` will failed with missing trienode in storage. We have some ways for reproducing like runing action SetState + CreateAccount + the other random action. Otherwise, we can make those test passed by setting `byzantium` for calling to Finalised without IntermediateRoot which will compute current root hash to update to state trie. Will note it for troubleshooting more.

core/state/statedb.go
```diff
@@ -514,12 +529,20 @@ func (s *StateDB) updateStateObject(obj *stateObject) {
 		s.setError(fmt.Errorf("updateStateObject (%x) error: %v", addr[:], err))
 	}
 
-	// If state snapshotting is active, cache the data til commit. Note, this
-	// update mechanism is not symmetric to the deletion, because whereas it is
-	// enough to track account updates at commit time, deletions need tracking
-	// at transaction boundary level to ensure we capture state clearing.
-	if s.snap != nil {
```
huyngopt1994, line 521: From myview, seem they expect that snap layer always enables when initing Statedb, so accounts here still points to snapAccounts as a cache layer, just renaming variable.

trie/trienode/node.go
```diff
@@ -186,9 +206,9 @@ func NewWithNodeSet(set *NodeSet) *MergedNodeSet {
 // Merge merges the provided dirty nodes of a trie into the set. The assumption
 // is held that no duplicated set belonging to the same trie will be merged twice.
 func (set *MergedNodeSet) Merge(other *NodeSet) error {
-	_, present := set.Sets[other.Owner]
+	subset, present := set.Sets[other.Owner]
 	if present {
-		return fmt.Errorf("duplicate trie for owner %#x", other.Owner)
+		return subset.Merge(other.Owner, other.Nodes)
```
huyngopt1994, line 211: hm Could u explain why they change the expected behaviour here?

trie/database.go
```diff
@@ -27,13 +27,17 @@ import (
...
+	"github.com/ethereum/go-ethereum/trie/triestate"
 )
 
 // Config defines all necessary options for database.
 type Config struct {
 	Cache     int    // Memory allowance (MB) to use for caching trie nodes in memory
 	Journal   string // Journal of clean cache to survive node restarts
 	Preimages bool   // Flag whether the preimage of trie key is recorded
+
+	// Testing hooks
```
huyngopt1994, line 39: what does it mean testing hooks? So it's only used when testing.

trie/trienode/node.go
Francesco4203: They did not explicitly explain that here. 
My intuition is that, although the current flow will never get to this error, the merge function should remain a merge function, that is, to merge NodeSets. Any unexpected behaviors should be from somewhere else and thus should be handled at those places. 
Therefore, it seems reasonable here. What do you think?

trie/database.go
Francesco4203: Yes. It's a callback function for testing.

core/state/statedb.go
Francesco4203: https://github.com/ethereum/go-ethereum/pull/26371

core/state/journal.go
Francesco4203: Added

trie/trienode/node.go
huyngopt1994: The new flow, should be accept the dirtied nodes change then overwrite it into nodeset with same owner with update flag tracking counter too, reject in case viloate owner (some issues here) It could be a general solution. Actually we don't run previous logic in real prod yet, so we can't make sure that `fmt.Errorf("duplicate trie for owner %#x", other.Owner)` can happen in real prod or not. But from my view, the new logic is quite cleary.

huyngopt1994:
LGTM 🙏 .

huyngopt1994 (approved):

PR #590, opened 2024-09-30
core/txpool/blobpool: use types.Sender instead of signer.Sender (#30473)

minh-bq:
commit https://github.com/ethereum/go-ethereum/commit/0a21cb4d21a254cc87c7ecc50462f8c2735a302f.

Use types.Sender(signer, tx) to utilize the transaction's sender cache and avoid repeated address recover.


andicrypt (approved):

andicrypt:
LGTM

PR #593, opened 2024-10-04
core, trie: Expose block number to statedb

Francesco4203:
References: https://github.com/ethereum/go-ethereum/pull/27753

Changes:
- clean up: `stateObject` already contains the db, so there is no need to pass the db in its methods
- `statedb.Commit()` now also pass the block number, which will be later used for PBSS to build history data


huyngopt1994:
LGTM

huyngopt1994 (approved):

PR #596, opened 2024-10-07
 rawdb: implement freezer resettable & state freezer 

huyngopt1994:
Reference: https://github.com/ethereum/go-ethereum/pull/26324

### DIFF
--- a/core/rawdb/freezer.go
+++ b/core/rawdb/freezer.go
@@ -270,46 +270,49 @@ func (f *Freezer) ModifyAncients(fn func(ethdb.AncientWriteOp) error) (writeSize
 }
 
 // TruncateHead discards any recent data above the provided threshold number, only keep the first items ancient data.
-func (f *Freezer) TruncateHead(items uint64) error {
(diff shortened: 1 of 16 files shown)


core/rawdb/freezer.go
```diff
@@ -270,46 +270,49 @@ func (f *Freezer) ModifyAncients(fn func(ethdb.AncientWriteOp) error) (writeSize
 }
 
 // TruncateHead discards any recent data above the provided threshold number, only keep the first items ancient data.
-func (f *Freezer) TruncateHead(items uint64) error {
+func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
```
Francesco4203, line 273: Can we also add descriptions for the returning old head number here? Just like for the TruncateTail.

Francesco4203 (approved):
LGTM

core/rawdb/freezer.go
huyngopt1994: oke let me update

PR #588, opened 2024-09-27
trie: add trie db wrapper; refactor trienode

Francesco4203:
References: 
- https://github.com/ethereum/go-ethereum/pull/27160 (only for moving memoryNode to NodeSet and rename) 
- https://github.com/ethereum/go-ethereum/pull/27176

Changes:
- Add wrapper for database since later, we will have another database for path-based. Note that the common features such as cache, preimage,... are in the wrapper interface.
- Move nodeset to trienode/node 
- Adjust tests: add scheme to tests for later testing on path-based
- Remove callback in db.Commit() since it is not used (ref https://github.com/ethereum/go-ethereum/pull/26637)

PR #603, opened 2024-10-15
all: reworkNodeResolver for working with multiple state schemes with …

huyngopt1994:
Continue PR https://github.com/axieinfinity/ronin/pull/588 with missing port changes for supporting path/hash scheme when using trienode iterator


Francesco4203 (approved):
LGTM

PR #607, opened 2024-10-18
core: add testnet create and precompile hook

minh-bq:
Early day in development, when developing for hardfork, some testnet blocks were created with development code and not compatible with the current master code. This commit adds testnet hooks to handle testnet bad blocks, makes the code compatible with testnet data.


huyngopt1994 (approved):

huyngopt1994:
Screenshot 2024-10-21 at 9 08 44 PM
Before apply in Testnet env with full sync mode

After applying 
Screenshot 2024-10-21 at 9 21 20 PM

PR #614, opened 2024-10-25
core, eth/downloader: pbss fix release v1.13.1

Francesco4203:
References:
- eth/downloader: prevent pivot moves after state commit https://github.com/ethereum/go-ethereum/pull/28126
- core, eth/downloader: fix genesis state missing due to state sync https://github.com/ethereum/go-ethereum/pull/28124


huyngopt1994 (commented):
LGTM

PR #620, opened 2024-10-30
cmd/ronin/chaincmd: open ancient freezer when init genesis

Francesco4203:
Before this PR, when initializing genesis, the ancient db is not opened for writing the genesis state, results in data inconsistency in the ancient db. Therefore, when processing the next blocks, the following error occurred
```
CRIT [10-30|04:53:29.422] Failed to truncate state history freezer err="out of range, tail: 0, head: 0, target: 90000"
```
This PR simply open the ancient db for writing genesis state when initializing it.


huyngopt1994:
just be clear, we don't pass this flag in entrypoint.sh because it will use the default value is ancient.

PR #624, opened 2024-10-31
[consortium-v1] Get list validators from genesis instead of statedb for snapsync.

huyngopt1994:
Related to recent PR merged to path-base implementing https://github.com/axieinfinity/ronin/pull/614
> State sync overwrites the persistent state, and this action can disrupt states that reference the current persistent state. The appropriate solution in such a situation is to mark the entire trie database as unusable, which prevents any subsequent state reads, and raises a layer stale error if the read operation is done by the prev-created stale state. The trie database will remain in this state until the state sync process is completed, and a usable persistent state is assembled.

In the coming enhancements related to avoid disruption genesis data  in the snap sync with path schem enable. In the Ronin implementing, we have some functions for getting Validators from StateDB during snap sync which violate this approach. 
```
	// set SCValidators function before initiating new miner to prevent miner starts without SCValidators
	if chainConfig.Consortium != nil {
		c := eth.engine.(*consortium.Consortium)
		stack.RegisterAPIs(c.APIs(eth.blockchain))
		c.SetGetSCValidatorsFn(func() ([]common.Address, error) {
			stateDb, err := eth.blockchain.State()
			if err != nil {
				log.Crit("Cannot get state of blockchain", "err", err)
				return nil, err
			}
			return state.GetSCValidators(stateDb), nil
		})
```
---
```
	if mode == SnapSync {
		if !d.snapSync {
			// Snap sync will directly modify the persistent state, making the entire
			// trie database unusable until the state is fully synced. To prevent any
			// subsequent state reads, explicitly disable the trie database and state
			// syncer is responsible to address and correct any state missing.
			if d.blockchain.TrieDB().Scheme() == rawdb.PathScheme {
				if err := d.blockchain.TrieDB().Disable(); err != nil {
					return err
				}
			}
```

### DIFF
--- a/consensus/consortium/v1/consortium.go
+++ b/consensus/consortium/v1/consortium.go
@@ -338,11 +338,10 @@ func (c *Consortium) snapshot(chain consensus.ChainHeaderReader, number uint64, 
 			if cpHeader != nil {
 				hash := cpHeader.Hash()
 
-				validators, err := c.getValidatorsFromContract(chain, number)
+				validators, err := c.getValidatorsFromGenesis()
 				if err != nil {
 					return nil, err
 				}
-
 				snap = newSnapshot(c.config, c.signatures, number, hash, validators)
 				if err := snap.store(c.db); err != nil {
 					return nil, err
@@ -756,6 +755,31 @@ func (c *Consortium) doCalcDifficulty(signer common.Address, number uint64, vali
 	return new(big.Int).Set(diffNoTurn)
 }
 
+// getValidatorsFromGenesis gets the list of validators from the genesis block support backward compatibility in v1, only used with Snap Sync.
+func (c *Consortium) getValidatorsFromGenesis() ([]common.Address, error) {
+	var validatorSet []string
+	switch {
+	case c.chainConfig.ChainID.Cmp(big.NewInt(2020)) == 0:
+		validatorSet = []string{
+			"0x000000000000000000000000f224beff587362a88d859e899d0d80c080e1e812",
+			"0x00000000000000000000000011360eacdedd59bc433afad4fc8f0417d1fbebab",
+			"0x00000000000000000000000070bb1fb41c8c42f6ddd53a708e2b82209495e455",
+		}
+	case c.chainConfig.ChainID.Cmp(big.NewInt(2021)) == 0:
+		validatorSet = []string{
+			"0x0000000000000000000000004a4bc674a97737376cfe990ae2fe0d2b6e738393",
+			"0x000000000000000000000000b6bc5bc0410773a3f86b1537ce7495c52e38f88b",
+		}
+	default:
+		return nil, errors.New("no validator set for this chain only support Mainnet & Testnet")
+	}
+	var addresses []common.Address
+	for _, str := range validatorSet {
+		addresses = append(addresses, common.HexToAddress(str))
+	}
+	return addresses, nil
+}
+
 // Read the validator list from contract
 func (c *Consortium) getValidatorsFromContract(chain consensus.ChainHeaderReader, number uint64) ([]common.Address, error) {
 	if chain.Config().IsFenix(big.NewInt(int64(number))) {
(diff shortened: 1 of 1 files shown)


consensus/consortium/v1/consortium.go
```diff
@@ -756,6 +774,35 @@ func (c *Consortium) doCalcDifficulty(signer common.Address, number uint64, vali
 	return new(big.Int).Set(diffNoTurn)
 }
 
+// getValidatorsFromGenesis gets the list of validators from the genesis block support backward compatibility in v1, only used with Snap Sync.
+func (c *Consortium) getValidatorsFromGenesis() ([]common.Address, error) {
+	if c.chainConfig.ConsortiumV2Block == nil {
```
minh-bq, line 779: What is the purpose of this check? ConsortiumV2Block is the hardfork block of version 2.

consensus/consortium/v1/consortium.go
```diff
@@ -756,6 +774,35 @@ func (c *Consortium) doCalcDifficulty(signer common.Address, number uint64, vali
...
+			"0x00000000000000000000000070bb1fb41c8c42f6ddd53a708e2b82209495e455",
+		}
+	case c.chainConfig.ChainID.Cmp(big.NewInt(2021)) == 0:
+		validatorSet = []string{
+			"0x0000000000000000000000004a4bc674a97737376cfe990ae2fe0d2b6e738393",
+			"0x000000000000000000000000b6bc5bc0410773a3f86b1537ce7495c52e38f88b",
+		}
+	default:
+		log.Info("chainID", c.chainConfig.ChainID)
+		return nil, errors.New("No validator set for this chain only support Mainnet & Testnet")
```
minh-bq, line 797: nit: please lowercase the first character in error

consensus/consortium/v1/consortium.go
```diff
@@ -338,9 +338,19 @@ func (c *Consortium) snapshot(chain consensus.ChainHeaderReader, number uint64,
 			if cpHeader != nil {
 				hash := cpHeader.Hash()
 
-				validators, err := c.getValidatorsFromContract(chain, number)
-				if err != nil {
-					return nil, err
+				var validators []common.Address
+				var err error
+				// Snap Sync mode.
+				if c.skipCheckpointHeaderCheck {
```
minh-bq, line 344: IMO, I don't think this is needed as getValidatorsFromGenesis must return the same as getValidatorsFromContract at genesis block so we can just use getValidatorsFromGenesis. It's okay to let it like this but I see that in getValidatorsFromLastCheckpoint there is no check like this but use getValidatorsFromGenesis directly. I think we should be consistent here.

consensus/consortium/v1/consortium.go
huyngopt1994: hm let me remove, this check is valid, maybe mis typo

consensus/consortium/v1/consortium.go
huyngopt1994: The first idea is try to limit the impact for snap sync, and keep the remaining logic when we don't use it. But in theory, those infor must be same. Hence, let me remove here for consistency.

consensus/consortium/v1/consortium.go
```diff
@@ -485,7 +484,8 @@ func (c *Consortium) Prepare(chain consensus.ChainHeaderReader, header *types.He
 	header.Extra = header.Extra[:extraVanity]
 
 	if number%c.config.Epoch == 0 {
-		validators, err := c.getValidatorsFromContract(chain, number)
+		var validators []common.Address
+		validators, err := c.getValidatorsFromGenesis()
```
minh-bq, line 488: This is incorrect. getValidatorsFromGenesis is only the same as getValidatorsFromContract at genesis block. The if case here is meant to be triggered every checkpoint. The Prepare is not important in sync process so I think we don't need to change this.

minh-bq:
nit: In the commit message, can we have commit title + newline + commit description? Right now we have commit title + commit description without newline in between.

consensus/consortium/v1/consortium.go
huyngopt1994: hmm, agree already tested again for verifying, seem prepare is only triggred when mining new block, so this logic will not happen when sync logic, so  I will keep the logic, only return data from genesis in snapshot and last checkpoint logic.

minh-bq (approved):

PR #632, opened 2024-11-19
consortium-v2: move extraData check with contract into a function

minh-bq:
This commit refactors the logic that checks extraData's validator information with contract data into a separate function. It also adds more information in the error cases.


andicrypt:
LGTM

andicrypt (approved):

PR #636, opened 2024-11-25
consortium-v2: only return when failing to verify extraData

minh-bq:
In commit 584db0f668 ("consortium-v2: move extraData check with contract into a function"), we refactor the code and mistakenly return even if we successfully verify extraData. Fix it by checking the error and only return if there is an error when verifying extraData.


huyngopt1994:
seem good, retested in PBSS.

huyngopt1994 (approved):

PR #638, opened 2024-12-02
consortium-v2: make the period definition consistent with contract

minh-bq:
Currently, Ronin detects the start of new period is the first block of an epoch
where that block's timestamp is on the different date from first block of
previous epoch. However, in contract, the start of new period is the first block
of epoch where the parent of it (the last block of previous epoch)'s timestamp
is on the different date from the last block of "previous previous" epoch. In
this commit, we change the Ronin logic to be consistent with contract after
Venoki hardfork.

### DIFF
--- a/consensus/consortium/v2/consortium_test.go
+++ b/consensus/consortium/v2/consortium_test.go
@@ -2255,28 +2255,12 @@ func TestIsPeriodBlock(t *testing.T) {
 		db:       db,
 		contract: mock,
 	}
-	validators := make([]common.Address, NUM_OF_VALIDATORS)
-	for i := 0; i < NUM_OF_VALIDATORS; i++ {
-		validators = append(validators, common.BigToAddress(big.NewInt(int64(i))))
-	}
-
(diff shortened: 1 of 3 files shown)


consensus/consortium/v2/consortium_test.go
```diff
@@ -2274,18 +2294,251 @@ func TestIsPeriodBlock(t *testing.T) {
...
+	isPeriodBlock, err = c.IsPeriodBlock(chain, header, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if isPeriodBlock {
 		t.Errorf("wrong period block")
 	}
 }
 
+func generateChain(
```
huyngopt1994, line 2317: Could we add some description/purpose for this method with flag `bumpTimeStampLastBlockOfEpoch` usage for setting the timestamp of  lastBlock

huyngopt1994 (approved):

chiphamskymavis (approved):

PR #639, opened 2024-12-03
core/vm: Add Venoki header's fields when verify double sign

chiphamskymavis:
After Venoki, these fields are added into block's header: BaseFee, BlodGasUsed, ExceedBlobGas.
We need to add them into the double sign verification process to make sure the logic up to date.

The separated struct due to lack off optional and null support of contract Abi

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -66,8 +66,67 @@ func (n *BlockNonce) UnmarshalText(input []byte) error {
 	return hexutil.UnmarshalFixedText("BlockNonce", input, n[:])
 }
 
-// BlockHeader is used for consortium precompiled contracts
-type BlockHeader struct {
+type BlockHeader interface {
+	GetNumber() *big.Int
+	GetChainId() *big.Int
+	GetExtraData() []byte
+	GetBenificiary() common.Address
+	ToHeader() *Header
+	Bytes(verifyHeadersAbi string, methodName string) ([]byte, error)
+}
+
+// FromHeader creates a BlockHeader from a Header, this due to the lack of null or optional fields in the abi package
+func FromHeader(header *Header, chainId *big.Int, isVenoki bool) BlockHeader {
+	if !isVenoki {
+		blockHeader := &BlockHeaderV1{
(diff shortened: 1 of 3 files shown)


core/types/block.go
```diff
@@ -98,20 +109,68 @@ func FromHeader(header *Header, chainId *big.Int) *BlockHeader {
...
+		BaseFee:     header.BaseFee,
 	}
 	copy(blockHeader.ParentHash[:], header.ParentHash.Bytes())
 	copy(blockHeader.StateRoot[:], header.Root.Bytes())
 	copy(blockHeader.TransactionsRoot[:], header.TxHash.Bytes())
 	copy(blockHeader.ReceiptsRoot[:], header.ReceiptHash.Bytes())
 	copy(blockHeader.MixHash[:], header.MixDigest.Bytes())
+	if header.BaseFee == nil {
+		blockHeader.BaseFee = nil
+	}
```
minh-bq, line 121: This is redundant, can be removed

core/vm/consortium_precompiled_contracts.go
```diff
@@ -516,20 +543,22 @@ func (c *consortiumVerifyHeaders) getSigner(header types.BlockHeader) (common.Ad
 	return signer, nil
 }
 
+// verify if two has same signers, same block, same consensus address, then it is double signed
+// return false if it is not double signed
```
minh-bq, line 547:
```suggestion
// verify returns true if 2 blocks has the same signer (consensus address), same block number
// but with different seal hash
```

minh-bq (commented):
I've just left some small comments. Overall, LGTM.

core/vm/consortium_precompiled_contracts.go
chiphamskymavis: Thank you, updated

core/types/block.go
chiphamskymavis: Oh, nice catch, updated this.

minh-bq (approved):

PR #641, opened 2024-12-09
internal/ethapi: implement eth_getBlockReceipts (#27702)

minh-bq:
commit https://github.com/ethereum/go-ethereum/commit/f1801a9feda8f81532c92077d2c9a8b785fd699b.

The response is the same as eth_getTransactionReceipt but allow user to get all receipts in a block instead of having to query each transaction as eth_getTransactionReceipt.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -863,6 +863,34 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, address common.A
 	return res[:], state.Error()
 }
 
+// GetBlockReceipts returns the block receipts for the given block hash or number or tag.
+func (s *PublicBlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]map[string]interface{}, error) {
+	block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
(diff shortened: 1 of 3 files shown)


internal/ethapi/api.go
```diff
@@ -1729,13 +1757,27 @@ func (s *PublicTransactionPoolAPI) GetTransactionReceipt(ctx context.Context, ha
 	// Derive the sender.
 	bigblock := new(big.Int).SetUint64(blockNumber)
 	signer := types.MakeSigner(s.b.ChainConfig(), bigblock)
+	return marshalReceipt(ctx, s.b, receipt, blockHash, blockNumber, signer, tx, int(index)), nil
```
huyngopt1994, line 1760: so it always return nil here for err ? 
In the previous logic 
```
		header, err := s.b.HeaderByHash(ctx, blockHash)
		if err != nil {
			return nil, err
		}
```
I still see it returns err when can't get HeaderbyHash

minh-bq: The blockHash is the result of this call: https://github.com/axieinfinity/ronin/pull/641/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R1744. So when reaching here, it's very unlikely we get the error. As you can see at that line, when that transaction does not exist or some error happens, it still returns nil in the error.

huyngopt1994: oke,  so expect the logic will be same with the above logic

huyngopt1994 (approved):

Issue #647, opened 2024-12-12
debug_traceBlockByNumber return is not compatible to geth


s7v7nislands:
#### System information

Geth version: `geth version`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

#### Expected behaviour

transactionHash should be txHash

```
curl --location 'https://xxxx' \
--header 'Content-Type: application/json' \
--data '[{
	"jsonrpc":"2.0",
	"method":"debug_traceBlockByNumber",
	"params":[
		"0x26d292f",
        {
            "tracer": "callTracer"
        }
	],
	"id":1
}]'

[
    {
        "jsonrpc": "2.0",
        "id": 1,
        "result": [
            {
                "txHash": "0x6c4df6dd68b9ab506494fb4525b86ff49f7371a2b16cbf217caca86bf02b06cf",
                "result": {
                    "from": "0x5ea0cebf694288d13a70e43ecc6df902c5b451db",
                    "gas": "0x7a120",
                    "gasUsed": "0x5721a",
                    "to": "0xfff9ce5f71ca6178d3beecedb61e7eff1602950e",
...
```

#### Actual behaviour

```
curl --location 'https://xxxx' \
--header 'Content-Type: application/json' \
--data '[{
	"jsonrpc":"2.0",
	"method":"debug_traceBlockByNumber",
	"params":[
		"0x26d292f",
        {
            "tracer": "callTracer"
        }
	],
	"id":1
}]'

[
    {
        "jsonrpc": "2.0",
        "id": 1,
        "result": [
            {
                "transactionHash": "0x6c4df6dd68b9ab506494fb4525b86ff49f7371a2b16cbf217caca86bf02b06cf",
                "result": {
                    "from": "0x5ea0cebf694288d13a70e43ecc6df902c5b451db",
                    "gas": "0x7a120",
                    "gasUsed": "0x5721a",
                    "to": "0xfff9ce5f71ca6178d3beecedb61e7eff1602950e",
...
```

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

minh-bq:
Thanks for your report.

I just want to know more about your use case. Do you use any libraries to use this method and this bug makes the library fail to parse the response?

s7v7nislands:
I will parse the traces and get the internal transactions like in ethscan internal transactoins. We support many evm compatible chains, and they return `txHash` in response.

PR #648, opened 2024-12-13
eth/tracers: make transaction trace response compatible with geth

minh-bq:
The current transactionHash field in transaction trace response should be named txHash to be compatible with go-ethereum. This commit adds txHash instead of replacing the current transactionHash field and make them have the same value. This can help to avoid breaking change for current users of these methods since users are unlikely to use strict DisallowUnknownFields when decoding the json response.

Fixes: #647


chiphamskymavis (approved):

trantienduchn:
lets wait util our clients are aware of

PR #651, opened 2024-12-19
consortium-v2: add parent list parameter to IsTrippEffective

minh-bq:
This is the same as commit b718105228d9 ("consortium-v2: add parent list parameter to IsPeriodBlock") but with the IsTrippEffective function. The parent list is passed to this function when it is called in the VerifyHeader path.

### DIFF
--- a/consensus/consortium/v2/consortium.go
+++ b/consensus/consortium/v2/consortium.go
@@ -359,7 +359,7 @@ func (c *Consortium) verifyFinalitySignatures(
 	header *types.Header,
 	parents []*types.Header,
 ) error {
-	isTrippEffective := c.IsTrippEffective(chain, header)
+	isTrippEffective := c.IsTrippEffective(chain, header, parents)
 	snap, err := c.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, parents)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


NganSM:
@minh-bq It seems like there are two methods named isTrippEffective (method [1](https://github.com/minh-bq/ronin/blob/afc8b68ae29fbc5d262bffe222c6a01215a3a50e/consensus/consortium/v2/snapshot.go#L170) and [2](https://github.com/minh-bq/ronin/blob/afc8b68ae29fbc5d262bffe222c6a01215a3a50e/consensus/consortium/v2/consortium.go#L2029)). The logic to determine if Tripp consensus rules are enabled is different for each method (method 1 determines by block number and method 2 determines by both period and block number).  Can you clarify this?

minh-bq:
The method 2 uses the period number in the snapshot, the method 1 is used in the snapshot itself so it cannot rely on the snapshot's period number, it must rely on the extraData of produced block which is filled in Prepare.

minh-bq:
Hmm, actually, we can implement method 1 to use same rule like method 2 without using snapshot information. The main problem is how to get the period number of last epoch. To get that we might need to traverse back the chain to get the header information, in case we don't use snapshot.

chiphamskymavis (approved):
LGTM

PR #659, opened 2025-01-15
core: don't prune blob sidecar in archive node

minh-bq:
This commit adds a flag --no-pruning-sidecar (default: true) to disable blob sidecar pruning in archive node. Blob sidecar is meant to be pruned, but in the early state where the offchain storage is not available, it might be better to keep it in archive node for debugging.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -225,6 +225,12 @@ var (
 		Value:    "full",
 		Category: flags.StateCategory,
 	}
+	NoPruningSideCarFlag = &cli.BoolFlag{
+		Name:     "no-pruning-sidecar",
+		Usage:    `Disable blob sidecar pruning in archive node`,
+		Value:    true,
+		Category: flags.StateCategory,
+	}
 	SnapshotFlag = &cli.BoolFlag{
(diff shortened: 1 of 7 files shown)


cmd/utils/flags.go
```diff
@@ -1907,6 +1913,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	if ctx.IsSet(GCModeFlag.Name) {
 		cfg.NoPruning = ctx.String(GCModeFlag.Name) == "archive"
 	}
+	if cfg.NoPruning && ctx.Bool(NoPruningSideCarFlag.Name) {
```
chiphamskymavis, line 1916: Feel like we should valid date this flag, something like: `No sidecar pruning is only enabling for archive mode`

chiphamskymavis (approved):
Look good overall

cmd/utils/flags.go
minh-bq: I've just added a warning

chiphamskymavis:
LGTM

PR #642, opened 2024-12-10
release: v2.9.0

huyngopt1994:
Including some changes 
* Path base implementing.
* Bugs/fixes during testing and auditing.

PR #660, opened 2025-01-16
core: ensure txindex will be triggered at least once (#27847)

chiphamskymavis:
Currently, we trigger the logic to (un)index transactions when the node receives a new block. However, in some cases the node may not receive new blocks (eg, when the Geth node is configured without peer discovery, or when it acts as an RPC node for historical-only data).

In these situations, the Geth node user may not have previously configured txlookuplimit (i.e. the default of around one year), but later realizes they need to index all historical blocks. However, adding txlookuplimit=0 and restarting geth has no effect. This change makes it check for required indexing work once, on startup, to fix the issue.

Cherry-pick: https://github.com/ethereum/go-ethereum/pull/27847

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2447,39 +2447,6 @@ func (bc *BlockChain) maintainTxIndex(ancients uint64) {
 		rawdb.IndexTransactions(bc.db, from, ancients, bc.quit)
 	}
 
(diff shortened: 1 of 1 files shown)


chiphamskymavis:
Wait for #642 merged

PR #661, opened 2025-01-16
account/abi: handle solidity panic revert

chiphamskymavis:
See https://docs.soliditylang.org/en/v0.8.21/control-structures.html#panic-via-assert-and-error-via-require


minh-bq (approved):
Please add the cherry-picked commit in go-ethereum in the commit message.

Overall, LGTM.

PR #662, opened 2025-01-21
eth/filters: enforce the limit of topics and sub-topics

minh-bq:
- eth/filters: exit early if topics-filter has more than 4 topics (#28494)

commit https://github.com/ethereum/go-ethereum/commit/e38b9f183065b374110918660114e93006f102dc.

Currently, geth's will return `[]` for any `len(topics) > 4` log filter. The
EVM only supports up to four logs, via LOG4 opcode, so larger criterias fail.
This change makes the filter query exit early in those cases.

- eth/filters: enforce topic-limit early on filter criterias (#29535)

commit https://github.com/ethereum/go-ethereum/commit/ef5ac3fb7ae5bf41a465cc32845631f01ff823ef.

This PR adds a limit of 1000 to the "inner" topics in a filter-criteria.

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -35,6 +35,13 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+var (
+	errExceedMaxTopics = errors.New("exceed max topics")
+)
+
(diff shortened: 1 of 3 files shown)


chiphamskymavis (approved):
LGTM

PR #676, opened 2025-02-03
accounts/abi: fix bigInt topic encoding

minh-bq:
- accounts/abi: fix bigInt topic encoding (#28764)

commit https://github.com/ethereum/go-ethereum/commit/d0edc5af4a2f4e8e9961c5da4d710579ff19681f.

Currectly encode the negative bigInt. Currently, bigInt(-1) is encoded as
0000000000...000000000001, while it should be ffffffffff...ffff instead.

- accounts/abi: fix MakeTopics mutation of big.Int inputs (#30785)

commit https://github.com/ethereum/go-ethereum/commit/3c754e2a09d112cd62f4d51f221637f9a147dcc6.

also changed the behavior of the function from just _reading_ the input
`*big.Int` via `Bytes()`, to leveraging `big.U256Bytes` which is documented as
being _destructive_:

This change updates `MakeTopics` to not mutate the original.

### DIFF
--- a/accounts/abi/topics.go
+++ b/accounts/abi/topics.go
@@ -42,7 +42,7 @@ func MakeTopics(query ...[]interface{}) ([][]common.Hash, error) {
 			case common.Address:
 				copy(topic[common.HashLength-common.AddressLength:], rule[:])
 			case *big.Int:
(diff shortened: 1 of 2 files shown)


chiphamskymavis (approved):
LGTM

PR #677, opened 2025-02-03
accounts/abi: return error on fixed bytes with size larger than 32 bytes (#26075)

minh-bq:
commit https://github.com/ethereum/go-ethereum/commit/8578eb2fe1071c70eeee1a76b14d87e93ba129ca.

* fixed bytes with size larger than 32 bytes is not allowed

* add testcase


chiphamskymavis (approved):
LGTM

PR #682, opened 2025-02-04
core/vm: remove consortiumLog precompiled contract after Cancun

minh-bq:
This precompiled contract is unused and may cause the node with environment variable DEBUG=true to halt. So we decide to remove this precompiled contract after Cancun and remove the logic when DEBUG=true, make this precompiled contract always return (input, nil).

Reported-by: Dontonka


chiphamskymavis (approved):
LGTM

PR #683, opened 2025-02-04
all: convert time.After to time.NewTimer to avoid leak

minh-bq:
In Go before 1.23, we need to call timer.Stop on the timer to allow it to be collected by garbage collector. So using the pattern like <-time.After(...) can lead to memory leak. Convert this pattern to time.NewTimer so that we can call Stop on created timer. This commit does not fix the pattern in tests and unused code (e.g. light, clique, ...)


minh-bq:
Closed because to fix the issue while maintaining the current behavior is complicated and there are breaking change in Go 1.23 in timer. So let's wait until we bump to Go 1.23 to fix the memory leak.

PR #684, opened 2025-02-04
core/vote: Prevent vote pool DOS

chiphamskymavis:


core/vote/vote_pool.go
```diff
@@ -156,12 +158,30 @@ func (pool *VotePool) putIntoVotePool(voteWithPeerInfo *voteWithPeer) bool {
...
+		isFutureVote = false
+		voteHash     = vote.Hash()
+	)
+
 	if _, ok := pool.originatedFrom[voteHash]; ok {
 		log.Debug("Vote pool already contained the same vote", "voteHash", voteHash)
 		return false
 	}
-	pool.originatedFrom[voteHash] = peer
+	defer func() {
```
minh-bq, line 171: I think it's better to move this logic right before we return true in this function than using defer

minh-bq (commented):
It's a quite interesting solution 👍

core/vote/vote_pool.go
chiphamskymavis: Nice catch, moved this logic to the end of this function.

core/vote/vote_pool.go
```diff
@@ -126,11 +126,7 @@ func (pool *VotePool) loop() {
 }
 
 func (pool *VotePool) PutVote(peer string, vote *types.VoteEnvelope) {
-	select {
-	case pool.votesCh <- &voteWithPeer{vote: vote, peer: peer}:
-	default:
-		log.Debug("Failed to put vote into vote pool")
-	}
+	pool.votesCh <- &voteWithPeer{vote: vote, peer: peer}
```
chiphamskymavis, line 129: Behavior change from non-blocking but drop message to wait for channel available

PR #691, opened 2025-02-13
txpool, evm: skip blacklist and whitelist deployer check after Venoki

chiphamskymavis:
After Venoki, we don't have to check for whitelisted and blacklisted contract deployers anymore.
This help boost performance by not access to state for checking.


minh-bq (commented):
Let's wait for confirmation from security team. Overall, LGTM.

minh-bq (approved):

PR #693, opened 2025-02-16
core/rawdb, p2p/discover/table.go, trie, core/state, eth/handler: cherry pick bug fixes for path base implementing & enr filtering

huyngopt1994:
This PR including some fixes related to path base implementing and ENR filtering feature.

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -190,9 +190,9 @@ func newHandler(config *handlerConfig) (*handler, error) {
 		}
 	} else {
 		head := h.chain.CurrentBlock()
-		if head.NumberU64() > 0 {
+		if head.NumberU64() > 0 && h.chain.HasState(head.Root()) {
 			// Print warning log if database is not empty to run fast sync.
-			log.Warn("Switch sync mode from fast sync to full sync")
+			log.Warn("Switch sync mode from fast sync to full sync", "reason", "snap sync complete")
 		} else {
 			// If fast sync was requested and our database is empty, grant it
 			h.fastSync = uint32(1)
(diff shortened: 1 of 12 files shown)


trie/trie_reader.go
```diff
@@ -51,6 +51,7 @@ type trieReader struct {
 
 // newTrieReader initializes the trie reader with the given node reader.
 func newTrieReader(stateRoot, owner common.Hash, db NodeReader) (*trieReader, error) {
+	// TODO: explain this logic
```
minh-bq, line 54: What does this line mean?

eth/handler.go
```diff
@@ -190,9 +190,9 @@ func newHandler(config *handlerConfig) (*handler, error) {
 		}
 	} else {
 		head := h.chain.CurrentBlock()
-		if head.NumberU64() > 0 {
+		if head.NumberU64() > 0 && h.chain.HasState(head.Root()) {
```
minh-bq, line 193: Can you add more description why this is not needed before but now it's needed (maybe in the comment or commit message)?

minh-bq (commented):
The first commit does not follow the normal commit convention. IMO, it should be named "trie: use EmptyRootHash instead of common.Hash{} for trie.NewEmpty".

I suggest adding something like "commit https://github.com/ethereum/go-ethereum/commit/193f350eb911c9d8a93577d619986e97f490b700." in the commit message of cherry-picked commit so that we can easily reference the information about these commits later.

trie/trie_reader.go
huyngopt1994: Seem it's unused comment, I removed and squad to the original commit.

eth/handler.go
huyngopt1994: Added.

